PR #4, opened 2015-06-15
Use the Host port when port forwarding is detected

weierophinney:
In vagrant and other systems, you often have a port on the host forwarding to a port on the client box. In terms of ServerUrl, the port on the host server is the public one that should be used.

Prior to this patch ServerUrl was detecting both the port from the Host header, as well as SERVER_PORT, and using *both* in generated URLs.

As an example, given:

- HTTP_HOST: localhost:10088
- SERVER_PORT: 10081

ServerUrl would generate the host "localhost:10088:10081", when it should be generating "localhost:10088". This patch corrects the situation.

### DIFF
--- a/src/Helper/ServerUrl.php
+++ b/src/Helper/ServerUrl.php
@@ -79,22 +79,23 @@ class ServerUrl extends AbstractHelper
 
         if (isset($_SERVER['HTTP_HOST']) && !empty($_SERVER['HTTP_HOST'])) {
             // Detect if the port is set in SERVER_PORT and included in HTTP_HOST
-            if (isset($_SERVER['SERVER_PORT'])) {
-                $portStr = ':' . $_SERVER['SERVER_PORT'];
-                if (substr($_SERVER['HTTP_HOST'], 0-strlen($portStr), strlen($portStr)) == $portStr) {
-                    $this->setHost(substr($_SERVER['HTTP_HOST'], 0, 0-strlen($portStr)));
+            if (isset($_SERVER['SERVER_PORT'])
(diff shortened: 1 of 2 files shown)


gianarb:
:+1:

src/Helper/ServerUrl.php
```diff
@@ -85,6 +85,16 @@ protected function detectHost()
                     $this->setHost(substr($_SERVER['HTTP_HOST'], 0, 0-strlen($portStr)));
                     return;
                 }
+
+                // If the Host header contains a port, and it differs from the
+                // SERVER_PORT, use the port from the Host header. Typically
+                // this is a situation where port-forwarding is in use, and you
+                // want to present a URI using the public port.
```
akrabat: Comment should be:
```
// If the Host header contains a port, then use this port. Typically
// this is a situation where port-forwarding is in use, and you
// want to present a URI using the public port.
```

or similar as we don't actually check that it differs from `SERVER_PORT` (or need to).

weierophinney: > or similar as we don't actually check that it differs from SERVER_PORT (or need to).

There's a block immediately before it that *does*, actually. :) This is the condition that we hit if the SERVER_PORT and the HTTP_HOST's port are not equal.

akrabat: Hmm... we should probably extract the port from HTTP_PORT once then :)

weierophinney:
As this was communicated from the Zend Server team, I'm also going to backport it to 2.4.3 as a critical fix for the LTS series.

PR #7, opened 2015-07-14
Updates in Helper\Navigation\Breadcrumbs.php and Menu.php

Zyqsempai:
Added possibility to set parameters(variables) for partial to populate in the view.
In the current version there is no possibility to transmit variables into partial view , you can only set partial and module like this:

```php
$this->navigation('navigation')
    ->menu()
    ->setPartial('partial/yourPartial.phtml','yourModule');
```

so with this update you can simply set those variables like this:

```php
$this->navigation('navigation')
    ->menu()
    ->setPartial('partial/yourPartial.phtml','yourModule')
    ->setPartialParams(array('variableName' =>$variable));
```

That's all.

### DIFF
--- a/src/Helper/Navigation/Breadcrumbs.php
+++ b/src/Helper/Navigation/Breadcrumbs.php
@@ -1,8 +1,10 @@
 <?php
+
 /**
- * Zend Framework (http://framework.zend.com/)
+ * Zend Framework (http://framework.zend.com/).
  *
  * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ *
  * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
@@ -15,49 +17,50 @@ use Zend\View;
 use Zend\View\Exception;
 
 /**
- * Helper for printing breadcrumbs
+ * Helper for printing breadcrumbs.
  */
 class Breadcrumbs extends AbstractHelper
 {
     /**
-     * Whether last page in breadcrumb should be hyperlinked
+     * Whether last page in breadcrumb should be hyperlinked.
      *
      * @var bool
      */
     protected $linkLast = false;
 
     /**
-     * The minimum depth a page must have to be included when rendering
+     * The minimum depth a page must have to be included when rendering.
      *
      * @var int
      */
     protected $minDepth = 1;
 
     /**
-     * Partial view script to use for rendering menu
+     * Partial view script to use for rendering menu.
      *
      * @var string|array
      */
     protected $partial;
 
     /**
-     * Breadcrumbs separator string
+     * Breadcrumbs separator string.
      *
      * @var string
      */
     protected $separator = ' &gt; ';
 
     /**
-     * Array with variables to populate in the view
+     * Array with variables to populate in the view.
      *
      * @var array
      */
     protected $partialParams = [];
 
     /**
-     * Helper entry point
+     * Helper entry point.
      *
-     * @param  string|AbstractContainer $container container to operate on
+     * @param string|AbstractContainer $container container to operate on
+     *
      * @return Breadcrumbs
      */
     public function __invoke($container = null)
@@ -70,12 +73,13 @@ class Breadcrumbs extends AbstractHelper
     }
 
     /**
-     * Renders helper
+     * Renders helper.
      *
      * Implements {@link HelperInterface::render()}.
      *
-     * @param  AbstractContainer $container [optional] container to render. Default is
-     *                                      to render the container registered in the helper.
+     * @param AbstractContainer $container [optional] container to render. Default is
+     *                                     to render the container registered in the helper.
+     *
      * @return string
      */
     public function render($container = null)
@@ -90,10 +94,11 @@ class Breadcrumbs extends AbstractHelper
 
     /**
      * Renders breadcrumbs by chaining 'a' elements with the separator
-     * registered in the helper
+     * registered in the helper.
      *
-     * @param  AbstractContainer $container [optional] container to render. Default is
-     *                                      to render the container registered in the helper.
+     * @param AbstractContainer $container [optional] container to render. Default is
+     *                                     to render the container registered in the helper.
+     *
      * @return string
      */
     public function renderStraight($container = null)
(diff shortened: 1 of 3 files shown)


froschdesign:
:-1:
This is not needed, because:

- [you can use menu options in partial view script](http://framework.zend.com/manual/current/en/modules/zend.navigation.view.helper.menu.html#using-menu-options-in-partial-view-script) (same for Breadcrumbs)
- or the view helper ["Placeholder"](http://framework.zend.com/manual/current/en/modules/zend.view.helpers.placeholder.html)

weierophinney:
@froschdesign From reading the description, I think the reporter wants to be able to pass arbitrary parameters to the partial, not just menu options.

@Zyqsempai Please add unit tests, and also run `./vendor/bin/php-cs-fixer fix` within your checkout before your next push.

froschdesign:
@weierophinney 
The description was not clear enough. Therefore, the second option: view helper "Placeholder".

froschdesign:
@Zyqsempai 
> ```php
->setPartial('partial/yourPartial.phtml','yourModule')
```

[There is no second parameter!](https://github.com/zendframework/zend-view/blob/master/src/Helper/Navigation/Menu.php#L674)

**Comment by froschdesign** [comment]
> Added possibility to set parameters(variables) for partial to populate in the view.

Can you provide a use case?

**Comment by Zyqsempai** [comment]
Hi , here is use case from my current project, 
we have few view's and we need to show on them one navigation-partial but with some differences , for that we need to transmit simple flag(show or not) into this partial with this update you can simple do like this:
$this->navigation('navigation')
    ->menu()
    ->setPartial('partial/yourPartial.phtml','yourModule')
    ->setPartialParams(array('flag' => false));

**Comment by Zyqsempai** [comment]
Added unit test for setter and getter for PartialParams , also fixed all issues with php-cs-fixer.

**Comment by froschdesign** [comment]
> for that we need to transmit simple flag(show or not)

Show or not? You can use the [custom page properties](http://framework.zend.com/manual/current/en/modules/zend.navigation.pages.html#custom-page-properties):

```php
$page1 = new Zend\Navigation\Page\Mvc();
$page1->header = true;

$page2 = new Zend\Navigation\Page\Mvc();
$page2->footer = true;
```
I'm not entirely against this PR, but I think there are too many options in the navigation view helpers.

> ```php
->setPartial('partial/yourPartial.phtml','yourModule')
```

Again: [there is no second parameter!](https://github.com/zendframework/zend-view/blob/master/src/Helper/Navigation/Menu.php#L674)

**Comment by weierophinney** [comment]
I like the general idea (passing variables to the partial). However, I'm not convinced by the implementation. Typically, I suspect I'd want to pass parameters to the partial on a case-by-case basis, at the time of invocation — in other words, I wouldn't want _stateful_ parameters. As such, I'm wondering if this should warrant an alternate rendering method instead of additional helper properties — something like this:

```php
echo $this->navigation()->menu()->renderPartialWithParams(['flag' => $flag]);
```

The signature would be:

```php
public function renderPartialWithParams(array $params = [], $container = null, $partial = null)
```

I've provided a "sketch" if the idea here: https://github.com/weierophinney/zend-view/commit/80912976f00a90ea1d7f84984edd74e424e07d13

@froschdesign — what do you think of this approach?

froschdesign:
@weierophinney 
Looks good to me. No extra options in the helper!

weierophinney:
@Zyqsempai Do you want to take it from here? or would you like me to do a competing PR?

Zyqsempai:
@weierophinney, What do you mean, "to take it from here"?
Regarding your idea i will be glad to help(if i can) to implement this functionality.

Martin-P:
Personally I think the method name indicates how it should be implemented: `renderPartialWithParams`. Render a partial with parameters, thus adding the parameters to the method `renderPartial`.

The suggested implementation makes me think about methods like `renderMenuWithContainer` or `renderMenuWithContainerAndOptions`. This is (fortunately) not implemented that way and instead there is a single method `renderMenu` with optional parameters `$container` and `$options`. It makes more sense to me to use the same approach with `renderPartial`.

weierophinney:
@Martin-P The problem is that `renderPartial()` has the following signature already:

```php
public function renderPartial($container = null, $partial = null)
```

These are both null by default, and the method does some work to auto-determine them if they are not present. Considering the standard use case is to pass *no* arguments, adding the parameters as an option to the end makes the method unwieldy:

```php
echo $this->navigation->menu()->renderPartial(null, null, ['flag' => $flag]);
```

This is why I suggested adding a new method, putting the argument at the start.

@Zyqsempai — by "take it from here," I was inviting you to re-factor your patch to follow the direction I was providing. :) I haven't written tests, as I just wanted to demonstrate how I'd approach it, so you'd need to:

- Merge my commit, or copy/paste it into your branch.
- Make similar changes for the Breadcrumbs helper.
- Write tests for each that validate that the parameters passed are being passed to the partial via the view model.

You can then push to this same pull request. I'd recommend starting from a fresh branch, and then force-pushing to the original branch on your github fork.

Zyqsempai:
@weierophinney , Thank you for your invitation, i will start it tomorrow.

Martin-P:
@weierophinney I understand your point, but isn't it more intuitive to keep the method more like the regular [partial view helper](https://github.com/zendframework/zend-view/blob/ef3d64b8815a069c60cf7b9a6de636ee26fc9fe0/src/Helper/Partial.php#L36)? At this moment when I want to use `renderPartial` with the default container I already have to use `null` as the first parameter.

Looking at `renderPartialWithParams` I think most people will use it like this:
```php
echo $this->navigation->menu()->renderPartialWithParams(
    ['flag' => $flag],
    null,
    'myPartial.phtml'
);
```
So perhaps it is better to ommit the `$container` completely here. We always have `setContainer` for this and don't really need it as a parameter(which only causes trouble). That way the method can be simplified:
```php
public function renderPartialWithParams($partial = null, array $params = [])
```
And looking at this... I can't ignore the possibility to introduce a BC break instead of creating new methods and change the signature for `renderPartial` into the same signature as the partial view helper:
```php
public function renderPartial($partial = null, array $params = [])
```

I am aware it is a BC break, but I think this is a better approach instead of creating fixes because of something that can possibly be called a design flaw?

weierophinney:
@Zyqsempai I'm trying to keep the order of the arguments *internally* consistent in the helpers. Since the `withPartial()` method already has container and partial in that order, it makes sense to keep them in that order throughout. We can revisit the parameter order at a later date.

Zyqsempai:
@weierophinney , So, what the plan? I will implement you idea, as discussed earlier? 
And also i have a question (sorry if this is stupid one but..) How i can get the ZF2 development version? I mean with all test and modules , doing it through composer i got latest ZF2 version but without tests, in other way by cloning modules one by one i don't have enough dependencies to start those tests. Also php-cs-fixer, i have installed manually. Those facts made me think that maybe i got "production" version or something like that. Thank you in advance.

froschdesign:
Please close this PR. It was merged with #8.

froschdesign:
ping @weierophinney

Issue #3, opened 2015-06-10
Zend view does not work standalone


Mech7:
Or at least is missing required dependencies: 

```
composer require zendframework zend-view
$view = new \Zend\View\Renderer\PhpRenderer();
```
Fatal error: Class 'Zend\ServiceManager\AbstractPluginManager' not found

The problem runs deeper then just view though now am trying to setup standalone forms and there are some issues with it too.

gianarb:
:+1: This component suggests `zendframework/zend-servicemanager` and this package is REQUIRED for your implementations..

Why is suggested and not required? :-1: There are a lot of response about this question..

@Mech7 you can try to add `zendframework\zend-servicemanager` in your composer.json.. :)

mtymek:
Sometime ago I blogged how to make it work standalone: http://mateusztymek.pl/blog/using-standalone-zend-view.
It's a bit quirky, but works.

weierophinney:
@Mech7 `Zend\View` is not *just* the PHP renderer. It also provides an event-driven view layer (which allows you to specify your own rendering and response strategies), and interfaces for things such as the renderers themselves. If you are doing JSON views, the JsonRenderer requires *no* extra dependencies whatsoever.

Our policy has been to only `require` dependencies that are used by *every* workflow, and to `suggest` all other dependencies. (For development purposes, most, if not all, suggested dependencies are required to allow testing.)

Now that the components are split, we'll also be going one step further. If you look at the work done on `Zend\Feed`, we created an interface for loading feed reader plugins, and a default implementation that specifies all the plugins we ship. When used as part of a full ZF2 application, however, the ServiceManager-based implementation is used instead, allowing for configurable plugins. We'll be doing similarly for the view helpers in this component as well.

However, right now, follow the *suggestions* shown when you install this component: it suggests that if you are using the PhpRenderer, you need to *also* install `zendframework/zend-servicemanager`. So install it.

PR #12, opened 2015-09-10
Add missing zend-filter dependency

codeliner:
While trying to use zend-view with zend-expressive I got an exception:

`Fatal error: Class 'Zend\Filter\FilterChain' not found in \proophessor-do\vendor\zendframework\zend-view\src\Renderer\PhpRenderer.php on line 419`

`Zend\Filter\FilterChain` is required by the `PhpRenderer` to work correctly so `zend-filter` should be required in composer.json.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
         "php": ">=5.5",
         "zendframework/zend-eventmanager": "~2.5",
         "zendframework/zend-loader": "~2.5",
-        "zendframework/zend-stdlib": "~2.5"
+        "zendframework/zend-stdlib": "~2.5",
+	"zendframework/zend-filter": "~2.5"
     },
     "require-dev": {
         "zendframework/zend-authentication": "~2.5",
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "zendframework/zend-eventmanager": "~2.5",
         "zendframework/zend-loader": "~2.5",
         "zendframework/zend-stdlib": "~2.5",
-	"zendframework/zend-filter": "~2.5"
+        "zendframework/zend-filter": "~2.5"
     },
     "require-dev": {
         "zendframework/zend-authentication": "~2.5",


composer.json
```diff
@@ -16,7 +16,8 @@
         "php": ">=5.5",
         "zendframework/zend-eventmanager": "~2.5",
         "zendframework/zend-loader": "~2.5",
-        "zendframework/zend-stdlib": "~2.5"
+        "zendframework/zend-stdlib": "~2.5",
+	"zendframework/zend-filter": "~2.5"
```
malukenho: miss align

maxgu:
+1 
Fatal error: Class 'Zend\Filter\FilterChain' not found in /storage/proj/school/vendor/t4web/form-renderer/vendor/zendframework/zend-view/src/Renderer/PhpRenderer.php on line 419

please add missing zend-filter dependency
or create PhpRenderer without FilterChain and PhpRendererFiltered renderer with FilterChain

danizord:
This is not required because people can use Zend\View without using PhpRenderer. See https://github.com/zendframework/zend-view/issues/3#issuecomment-111167149

codeliner:
@danizord thx for the info. IMHO this is a bad approach. Zend\Filter is used internally by the PhpRenderer. It is no dependency I have to set up as a user. So how should I now about it? Should I require all the suggestions myself just to make sure that nothing breaks? Oh or the other way would be to study the internals first and find at line [419](https://github.com/zendframework/zend-view/blob/master/src/Renderer/PhpRenderer.php#L419) of the PhpRenderer the hint that zend-filter is suggested because the PhpRenderer assumes that it is available? 

IMHO we have two solutions here: 
1. Add zend-filter as dependency
2. Refactor PhpRenderer to not produce a fatal error

But we can also just close the issue. Have a lot of other stuff on my desk.

maxgu:
@danizord in our project i use only PhpRenderer from Zend/View.

ZF2 - is a great library - It small pieces (like Zend\View) i may use separately in different our projects. And i use it, i want use it and not disturb about dependencies, i don't want coding this functionality again and again in our different projects, i want use ZF2 libs. It tested, flexible and clear. But i don't want review all code for search broken dependency. I want download and use it (and i expect to work from the box)

froschdesign:
@codeliner
> 1. Add zend-filter as dependency

:+1: 

> But we can also just close the issue.

This cannot be an option.

Maks3w:
I'll likely reject this because:

* This topic has been discussed lots of times and the result is if we add every dependency then the result is download the whole framework.

* Zend Expressive now require `zend-filter` when zend view is used.

codeliner:
@Maks3w :( If the components assume that the whole framework is installed. well, then the whole framework must be installed or the components need to be refactored. IMO a bad decision to just close the issue. But it is your choice ...

Maks3w:
We alredy know most components has too many suggestions and may indicate the component should be refactored and splitted. Anyway the conclusion the suggested dependencies won't be added to required.

PR #14, opened 2015-09-14
Added possibility change a template of menu item

bogutskyy:
Pull request from issue 13. Added possibility set menu items template in menu helper
```
$this->navigation()->menu()
->setItemTemplate('{itemOpenTag}{icon}{itemLabel}{itemCloseTag}')
->setItemPlaceholders(['{icon}' => '<i class="user-icon"></i>'])
```
or in pages properties ``menu_item_template`` and ``menu_item_placeholders``.


froschdesign:
I see some problems:

1. To modify the HTML you can use already a partial view script
2. …or XSLT.
3. There is no need to inject an element for an icon. Use simple CSS: (short example)

```html
<li class="glyphicon glyphicon-asterisk">…</li>
```
```css
.glyphicon {
    font-family: 'Glyphicons Halflings';
}

.glyphicon-asterisk:before {
    content: "\2a";
    /* more styling */
}
```

4. Your implementation is only for the menu helper.
5. For ZF3 it is the wrong, because we need simplification and flexibility at the same time – your solution only adds more code.

weierophinney:
I agree with @froschdesign here. There are already mechanisms for accomplishing this behavior, and adding the functionality per this patch adds to the maintenance overhead, as well as educational overhead.

Thanks for submitting, however, and looking forward to further submissions in the future!

PR #15, opened 2015-09-30
Updated zend-view to eventmanager v3

weierophinney:
- Uses dev-develop until we tag EM3.
- Updates trigger and attach calls to follow v3.
- Updates tests:
  - strategy tests primarily needed tooling for doing assertions on listeners and priorities after attaching aggregates (and updating aggregates to attach using the new signature).
  - Navigation and UrlIntegration tests were marked as incomplete as they depend on zend-mvc for testing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     },
     "require": {
         "php": ">=5.5",
-        "zendframework/zend-eventmanager": "~2.5",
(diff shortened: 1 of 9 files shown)


weierophinney:
Test coverage is expected, due to marking the navigation and URL integration tests as incomplete. However, this needs to be on the develop branch ASAP so we can continue with the zend-mvc updates. We can re-enable them once we have working code on the zend-mvc develop branch.

PR #17, opened 2015-10-05
Updated zend-view to zend-servicemanager v3

weierophinney:
Updates zend-view to use the new zend-servicemanager v3 API. Changes include:

- Removed `ServiceLocatorAwareInterface` implementation from `FlashMessenger` view helper; was never used internally, and the interface no longer exists anyways.
- Navigation `AbstractHelper` now removes the `ServiceLocatorAwareInterface`, but keeps the methods originally defined by it, using duck-typing to do injection when necessary.  Navigation view helpers require pulling navigation containers from the application container instance. This is automated by the Navigation `PluginManager` implementation now.
- All factories were updated to use the new `FactoryInterface` definition from zend-servicemanager:
  - New namespace for the `FactoryInterface`.
  - New method name for the factory (`__invoke`).
  - New arguments for the factory (first argument is a container-interop `ContainerInterface`; required second argument `$name` and optional third argument `$options`).
  - Factories that pulled the parent locator were updated, as the container passed is now the application service manager.
- Plugin managers were updated to the new `AbstractPluginManager` implementation:
  - Define `$instanceOf` and remove `validatePlugin` when applicable.
  - Define default configuration, merge it with incoming configuration, and pass to the parent constructor.
  - Define copious aliases to cover various naming patterns previously covered by normalization.
  - Initializers now have a new signature; all plugin-defined initializers were updated to it.
- When lazy loading plugin managers, pass an empty `ServiceManager` instance.
- All tests were updated to ensure plugin and service managers are properly injected and configured.

### DIFF
--- a/src/HelperPluginManager.php
+++ b/src/HelperPluginManager.php
@@ -9,9 +9,10 @@
 
 namespace Zend\View;
 
+use Interop\Container\ContainerInterface;
 use Zend\I18n\Translator\TranslatorAwareInterface;
 use Zend\ServiceManager\AbstractPluginManager;
-use Zend\ServiceManager\ConfigInterface;
+use Zend\Stdlib\ArrayUtils;
 
 /**
  * Plugin manager implementation for view helpers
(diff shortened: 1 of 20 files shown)


weierophinney:
Less test coverage is expected, as a number of tests had to be disabled until the collaborate components are updated to SM v3.

src/HelperPluginManager.php
```diff
@@ -22,59 +23,117 @@
...
-        'partial'             => 'Zend\View\Helper\Partial',
-        'placeholder'         => 'Zend\View\Helper\Placeholder',
-        'renderchildmodel'    => 'Zend\View\Helper\RenderChildModel',
-        'rendertoplaceholder' => 'Zend\View\Helper\RenderToPlaceholder',
-        'serverurl'           => 'Zend\View\Helper\ServerUrl',
-        'url'                 => 'Zend\View\Helper\Url',
-        'viewmodel'           => 'Zend\View\Helper\ViewModel',
+    protected $instanceOf = Helper\HelperInterface::class;
+
+    protected $config = [
```
mwillbanks: Can we not just force this to be a single way rather than alias other options?  I know it makes the BC break a bit more painful but it would be nice if it was always a single way.

weierophinney: There's a method behind the madness.

You can define keys up-front if you will for a number of the services. However, one of them in particular is not supported: invokables. The reason is because internally, `configure()` takes the `invokables` key and creates aliases and factories (pointing a class to the `InvokableFactory`) out of them. As such, these, in particular, need to be merged with incoming configuration and passed to the constructor *anyways*.

Additionally, there's another reason: Internally, the `Navigation\PluginManager` implementation extends `HelperPluginManager`. This is done to automate injection of the renderer and/or translator into the resolved helper on instantiation. However, that plugin manager *does not* and should not use the same configuration as the view helper plugin manager, as it should only resolve the various navigation helpers. As such, having these as a single array makes the inheritance story far easier.

Issue #20, opened 2015-12-01
HeadScript: handle empty type


matbech:
The HeadScript view helper should handle an empty type. In HTML5 the type attribute is optional. It would be nice if we can pass an empty type to the append* methods which then results in the type attribute being omitted in the output.

```
function itemToString
...
        $type = ($this->autoEscape) ? $this->escape($item->type) : $item->type;
        $html = '<script type="' . $type . '"' . $attrString . '>';
```

froschdesign:
Fixed with #21

PR #21, opened 2015-12-01
HeadScript: Handle empty type for HTML5

matbech:
Handle empty type as described in issue https://github.com/zendframework/zend-view/issues/20


froschdesign:
@matbech 
Thanks for the PR.

Two problems:

- no unit tests
- check for doctype is missing ([XHTML needs the type attribute](http://www.w3.org/TR/2003/WD-xhtml2-20030506/mod-scripting.html#sec_16.2.3.))

matbech:
@froschdesign done. I ended up using the special case only if a view is available and the doctype is set to HTML5.

PR #22, opened 2015-12-11
Added optional attribute 'async'

FabianKoestring:
Added `async` to optional attributes for executing scripts asynchronously.
http://www.w3.org/TR/html5/scripting-1.html


froschdesign:
@FabianKoestring 
Please add an unit test for the new attribute.

FabianKoestring:
@froschdesign 
Added!

PR #24, opened 2015-12-22
Zend filter is a hard dependency

harikt:
When calling `PhpRenderer` it cannot call `render` method for the https://github.com/zendframework/zend-view/blob/859542781565d582bb6d552f9a2ac8bca598ff61/src/Renderer/PhpRenderer.php#L519 have dependency on the `Filter` . The constructor doesn’t specifically mentions the Filter class is needed and throws error . https://github.com/zendframework/zend-view/blob/859542781565d582bb6d552f9a2ac8bca598ff61/src/Renderer/PhpRenderer.php#L145 .


danizord:
`Zend\Filter` is not required when not using `PhpRenderer`. See https://github.com/zendframework/zend-view/pull/12 and https://github.com/zendframework/zend-view/issues/3#issuecomment-111167149

weierophinney:
Closing for the rationale @danizord has stated.

harikt:
To be frank I disagree with the statements expressed in the issues. It just brings frustration when people trying to integrate things. Sorry to say it.

weierophinney:
@harikt I get your frustration; there's a reason the issue has been raised a number of times.

We're in an awkward position: if we don't have the dependency, we have some folks upset that it "does not work out of the box." If we add the dependency, we have some folks upset that they're forced to install a dependency they aren't using. We've chosen to err on the side of not providing as many dependencies, as we can document, "to use this, you must also require this in your project;" we cannot document removing a dependency, as it's not possible.

danizord:
@harikt I think #25 fixes your issue.

PR #25, opened 2015-12-22
Avoid needless Zend\Filter dependency

danizord:
As far I know the filter chain of `PhpRenderer` is not used by `Zend\Mvc` default rendering strategy (@weierophinney can you confirm?) and it seems like most users of `PhpRenderer` does not use the filter chain feature at all.

So I added this check to avoid `Zend\Filter` dependency when the filter chain is not explicity defined by `setFilterChain()` nor lazy initialized by `getFilterChain()`.

Maybe that fixes https://github.com/zendframework/zend-view/issues/3, https://github.com/zendframework/zend-view/pull/12 and https://github.com/zendframework/zend-view/pull/24.

### DIFF
--- a/src/Renderer/PhpRenderer.php
+++ b/src/Renderer/PhpRenderer.php
@@ -515,7 +515,11 @@ class PhpRenderer implements Renderer, TreeRendererInterface
 
         $this->setVars(array_pop($this->__varsCache));
 
-        return $this->getFilterChain()->filter($this->__content); // filter output
+        if ($this->__filterChain instanceof FilterChain) {
+            return $this->__filterChain->filter($this->__content); // filter output
+        }
+
+        return $this->__content;
     }
 
     /**
--- a/test/PhpRendererTest.php
+++ b/test/PhpRendererTest.php
@@ -442,4 +442,14 @@ class PhpRendererTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals(2, $this->renderer->sharedinstance());
         $this->assertEquals(3, $this->renderer->sharedinstance());
     }
+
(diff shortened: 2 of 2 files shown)


codeliner:
:+1:

harikt:
Hi @danizord , 

Can I understand what the filter chain exactly is doing ?

My feelings are in this direction for I fear magic behind this. I am sure people will come blame if they expect something from the Filter .

1. Specifically mention about the dependency on composer installation as @weierophinney him self has mentioned (https://github.com/zendframework/zend-view/pull/24#issuecomment-166697928) .

```
composer require zendframework/zend-view zendframework/zend-filter
```

1. These ( https://github.com/zendframework/zend-view/pull/24/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L46 ) sort of suggest is just useless for it don't say if you are using phprenderer you need this component also. What does just suggest with component name says?

So I am not in favour of this PR for I don't know the drawbacks if the filter is not calling.

danizord:
> So I am not in favour of this PR for I don't know the drawbacks if the filter is not calling.

No drawbacks. When an instance of `FilterChain` is present the behavior keeps the same. This PR only changes the behavior when there is no `FilterChain` defined.

Previously it was creating an empty `FilterChain` that actually does nothing, now it is just skipping.

harikt:
ok cool. Thanks for the information @danizord . 

So what is the real purpose of keeping the unused `FilterChain` object then?

danizord:
>So what is the real purpose of keeping the unused FilterChain object then?

Mainly for backwards compatibility. But even if it is not widely used, it is still useful if you want output filtering.

danizord:
Pinging @weierophinney for review :)

PR #34, opened 2016-02-08
Remove stream wrapper functionality

### LINKED ISSUE #23: [v3] remove Zend\View\Stream and useStreamWrapper  option
If I don't misunderstood, zf1 Zend_View_Stream was introduced for short_open_tag off environment , 

but `Since PHP 5.4.0, <?= is always available.`
http://php.net/manual/en/ini.core.php#ini.short-open-tag

so I think Zend_View_Stream can be removed since zf3.

mtymek:
Fixes #23.


mtymek:
Tests are failing, but it looks like compatibility issues with other ServiceManager v3.

sasezaki:
As I said on #23,
`Since PHP 5.4.0, <?= is always available.`
Is this BC break ?
( situation when user uses extending \Zend\View\Stream ?)

Ocramius:
@sasezaki the reason this is labeled as BC break is removal of API (even if protected). The correct approach could be deprecation + removal of these method bodies, while the API survives till the next major release.

sasezaki:
@Ocramius thanks describe reason

Ocramius:
@sasezaki it's an API change, which means that any subclasses may be broken by the change (that's why we use `private` by default now)

weierophinney:
This repository has been closed and moved to laminas/laminas-view; a new issue has been opened at https://github.com/laminas/laminas-view/issues/28.

weierophinney:

PR #35, opened 2016-02-09
[WIP] V3 readiness

mtymek:
This PR prepares `zend-view` for supporting ServiceManager v3.

* [x] update repository according to [migration/compatibility guide](https://github.com/zendframework/maintainers/wiki/ZF3-ServiceManager-component-refactors,-phase-2)
* [x] fix failing tests for SMv2
* [x] fix failing tests for SMv3
* [x] ensure everything works together after `zend-mvc` is v3-compatible

Last element is blocked by zendframework/zend-cache#68 - SMv3 cannot be installed at this moment because cache component is not V3-ready.

Edit: `zend-cache` was fixed, apparently `zend-mvc` is the blocker now.


test/HelperPluginManagerTest.php
```diff
@@ -55,10 +56,10 @@ public function testNoRendererInjectedInHelperWhenRendererIsNotPresent()
 
     public function testRegisteringInvalidHelperRaisesException()
     {
-        $helpers = new HelperPluginManager(new ServiceManager(), ['services' => [
-            'test' => $this,
+        $helpers = new HelperPluginManager(new ServiceManager(), ['factories' => [
+            'test' => function () { return $this; },
         ]]);
-        $this->setExpectedException('Zend\ServiceManager\Exception\InvalidServiceException');
+        $this->setExpectedException('Zend\View\Exception\InvalidHelperException');
```
kynx: This will need to check for a different exception depending on whether it's run under v2 or v3. Easiest way is to call a `getInvalidServiceException()` method in the same way the `CommonPluginManagerTrait` does. Or, if it's not a whole load of retooling, use the trait in this test and ditch the compatibility test.

test/HelperPluginManagerTest.php
```diff
@@ -67,7 +68,7 @@ public function testLoadingInvalidHelperRaisesException()
         $helpers = new HelperPluginManager(new ServiceManager(), ['invokables' => [
             'test' => get_class($this),
         ]]);
-        $this->setExpectedException('Zend\ServiceManager\Exception\InvalidServiceException');
+        $this->setExpectedException('Zend\View\Exception\InvalidHelperException');
```
kynx: Same here

src/Helper/Service/IdentityFactory.php
```diff
@@ -25,10 +26,22 @@ class IdentityFactory implements FactoryInterface
      */
     public function __invoke(ContainerInterface $container, $name, array $options = null)
     {
+        $container = $container->getServiceLocator();
```
kynx: This will throw an `E_USER_DEPRECATED` in v3. Is it really needed though? AFAIK in v3 this should be called with the `creationContext` of the container - i.e. the "root" ServiceManager if called from a plugin manager.

But I'm guessing you added it for a reason. If you need to distinguish between v2 and v3 the only real solution I've come up with is duck-typing the `configure()` method (which is only in v3). Ugly, but it works...

```
if (! method_exists($container, 'configure')) {
    $container = $container->getServiceLocator();
}
```

composer.json
```diff
@@ -36,7 +36,7 @@
         "zendframework/zend-paginator": "~2.5",
         "zendframework/zend-permissions-acl": "~2.5",
         "zendframework/zend-serializer": "~2.5",
-        "zendframework/zend-servicemanager": "dev-develop as 2.7.0",
+        "zendframework/zend-servicemanager": "^2.7.4 || ^3.0.3",
```
kynx: Should be `^2.7.5 || ^3.0.3`

src/Helper/Service/FlashMessengerFactory.php
```diff
@@ -25,6 +26,7 @@ class FlashMessengerFactory implements FactoryInterface
      */
     public function __invoke(ContainerInterface $container, $name, array $options = null)
     {
+        $container               = $container->getServiceLocator();
```
kynx: See comment on IdentityFactory

kynx:
Good stuff! Don't forget to add this to https://github.com/zendframework/maintainers/wiki/ZF3-ServiceManager-component-refactors,-phase-2

src/Helper/Service/IdentityFactory.php
mtymek: Thanks for the tip!

Yes, I'd like to make it work with both v2 and v3, so this will be helpful.

mtymek:
I updated the code according to kynx's feedback. I tested it with ServiceManager v3 (by temporarily setting required version as `3.0.3 as 2.8.0`), and then I saw only 7 tests failing, apparently all related to `zend-mvc` being not compatible yet - apparently it is the last blocker.
I will wait until it is ready, and then do the final check.

weierophinney:
@mtymek The approach I've been using for dependencies that are in the `require-dev` section, but which do not yet have forwards-compat releases, is as follows:

- When testing on travis-ci against servicemanager and/or eventmanager v3, *remove* the dependency (`composer remove --no-update zendframework/zend-such-and-such`).
- Find tests related to the component, and mark them skipped, with a message indicating they should be re-enabled once the component has forwards-compat releases.

This works in most situations, particularly when we start seeing circular dependencies (essentially, *any* zend-mvc dependency observed is likely a circular dependency at this point).

In this particular component, we have forwards-compat changes for a good portion of the components listed in the `require-dev` section, and all items in `require` have stable forwards-compat releases. If the only tests failing currently are due to zend-mvc, we can safely use this approach.

I'm willing to do that work today when merging; zend-view is a dep for a number of other components (navigation and form in particular), so this would help unblock some of those.

src/HelperPluginManager.php
```diff
@@ -143,41 +147,78 @@ class HelperPluginManager extends AbstractPluginManager
...
         Helper\BasePath::class            => InvokableFactory::class,
+        'zendviewhelperbasepath'          => InvokableFactory::class,
         Helper\Cycle::class               => InvokableFactory::class,
+        'zendviewhelpercycle'             => InvokableFactory::class,
         Helper\DeclareVars::class         => InvokableFactory::class,
+        'zendviewhelperdeclarevars'       => InvokableFactory::class,
         Helper\Doctype::class             => InvokableFactory::class, // overridden by a factory in ViewHelperManagerFactory
+        'zendviewhelperdoctype'           => InvokableFactory::class,
         Helper\EscapeHtml::class          => InvokableFactory::class,
+        'zendviewhelperescapehtml'        => InvokableFactory::class,
```
weierophinney: Separate the FQCN and normalized names, and group them by type. This will allow us to remove the normalized versions more easily at a later date. (I can do this on merge if needed.)

test/Helper/FlashMessengerTest.php
```diff
@@ -54,27 +55,28 @@ public function seedCurrentMessages()
...
-                            'flashmessenger' => 'Zend\View\Helper\Service\FlashMessengerFactory',
-                        ],
-                    ]);
-                },
-            ],
-        ]);
+        $config = new Config(
+            [
+                'services' => [
+                    'Config' => $config,
```
weierophinney: Use lowercase `config` (that's what v2 normalizes to, and ensures forwards compat). (I can do this on merge if needed.)

mtymek:
* separated FQCN from normalized name
* switched "config" to lowercase

mtymek:
As per your suggestion, I marked some tests as "skipped". Unfortunately I still get some errors in `PluginManagerCompatibilityTest` under v3, that I'm not able to resolve:

```
1) ZendTest\View\PluginManagerCompatibilityTest::testPluginAliasesResolve with data set #25 ('flashmessenger', 'Zend\View\Helper\FlashMessenger')
Zend\ServiceManager\Exception\ServiceNotCreatedException: Service with name "Zend\View\Helper\FlashMessenger" could not be created. Reason: Zend\Mvc\Controller\PluginManager::validatePlugin() has been deprecated as of 3.0; please define validate() instead

/home/mat/public_html/zend-view/vendor/zendframework/zend-servicemanager/src/ServiceManager.php:707
/home/mat/public_html/zend-view/vendor/zendframework/zend-servicemanager/src/ServiceManager.php:196
/home/mat/public_html/zend-view/vendor/zendframework/zend-servicemanager/src/AbstractPluginManager.php:143
/home/mat/public_html/zend-view/vendor/zendframework/zend-servicemanager/src/Test/CommonPluginManagerTrait.php:75

Caused by
PHPUnit_Framework_Error_Deprecated: Zend\Mvc\Controller\PluginManager::validatePlugin() has been deprecated as of 3.0; please define validate() instead

/home/mat/public_html/zend-view/vendor/zendframework/zend-servicemanager/src/AbstractPluginManager.php:157
/home/mat/public_html/zend-view/vendor/zendframework/zend-servicemanager/src/AbstractPluginManager.php:144
/home/mat/public_html/zend-view/vendor/zendframework/zend-mvc/src/Controller/PluginManager.php:84
/home/mat/public_html/zend-view/src/Helper/Service/FlashMessengerFactory.php:35
/home/mat/public_html/zend-view/vendor/zendframework/zend-servicemanager/src/ServiceManager.php:700
/home/mat/public_html/zend-view/vendor/zendframework/zend-servicemanager/src/ServiceManager.php:196
/home/mat/public_html/zend-view/vendor/zendframework/zend-servicemanager/src/AbstractPluginManager.php:143
/home/mat/public_html/zend-view/vendor/zendframework/zend-servicemanager/src/Test/CommonPluginManagerTrait.php:75
```

How can I work around it?

weierophinney:
@mtymek ha! I've already pulled locally to review!

In that particular case, the plugin requires zend-session, which does not have a forwards-compat version yet. As such, I'd override the `aliasProvider()` method and have it skip that alias; you can see an example of that in the [FilterPluginManagerCompatibilityTest](https://github.com/zendframework/zend-filter/blob/master/test/FilterPluginManagerCompatibilityTest.php#L39-L59).

Since I'm in the process of review anyways, I can do that for you if you'd like.

mtymek:
OK, please do - I was about to finish for today :-)

weierophinney:
Merged with #42!

PR #36, opened 2016-02-09
[Docs] Fixes #28 - Check Documentation Code Blocks

### LINKED ISSUE #28: Check Documentation Code Blocks
# Check code blocks are correct

> TLDR; Check in all files that codeblocks are correct, in PSR-2 format and have PHP syntax highlighting applied.

Code blocks should be in the following format...

<pre>
```php
<?php
use Zend\Barcode\Barcode;

// Only the text to draw is required
$barcodeOptions = array('text' => 'ZEND-FRAMEWORK');

// No required options
$rendererOptions = array();
$renderer = Barcode::factory(
    'code39', 'image', $barcodeOptions, $rendererOptions
);

```
</pre>

Note the three backticks then *php* in the opening fence, and the closing fence is just three backticks. It's common for the opening backticks to have no code type, or something like *source*.

Code should also have been automatically formatted into PSR-2 format, but sometimes these slip through the net.

froschdesign:


gianarb:
:+1:

PR #37, opened 2016-02-09
[Docs] Fixes #27 - Check All Headers In Documentation

### LINKED ISSUE #27: Check All Headers In Documentation
# Check headers are correct

> TLDR; Headers should use the ```#```, ```##``` etc to format different levels of headers, and not be underlines using ```=====``` or ```-----``, or be psuedo header using bold

Check all headers on the documentation - headers should use the hash style of declaration rather then be underlined with equals or dashes. The more hashes, the more of a subheading. Eg:

* ```#``` is equal to ```<h1>```
* ```##``` is equal to ```<h2>```
* ```###``` is equal to ```<h3>```
* ```####``` is equal to ```<h4>```
* ```#####``` is equal to ```<h5>```

Headings should be appropriate for their level in the documentation.

Psuedo headers using bold tags ```**``` should be replaced with appropriate level of heading tag.

froschdesign:


gianarb:
:+1:

Issue #46, opened 2016-02-18
Missing v2 cannonical FQSNs in Zend\View\Helper\Navigation\PluginManager.php (2.6.0 release)


webimpress:
For backward compatibility we are missing in https://github.com/zendframework/zend-view/blob/master/src/Helper/Navigation/PluginManager.php#L47-L52 v2 canonical FQCNs:
```
    'zendviewhelpernavigationcreadcrumbs' => InvokableFactory::class,
    'zendviewhelpernavigationlinks' => InvokableFactory::class,
    'zendviewhelpernavigationmenu' => InvokableFactory::class,
    'zendviewhelpernavigationsitemap' => InvokableFactory::class,
```

as we have in: https://github.com/zendframework/zend-view/blob/master/src/HelperPluginManager.php#L186

(It was noticed by @tmcnicholls in https://github.com/zendframework/zend-view/issues/45#issuecomment-185736440)

weierophinney:
I'll address this in #47; adding some code for this now.

PR #47, opened 2016-02-18
Ensure Navigation helper PluginManager is BC

weierophinney:
Adds tests similar to those for the `HelperPluginManager` to ensure that the first argument to the `Navigation\PluginManager` is optional and/or accepts a `Zend\ServiceManager\Config` instance, and updates the constructor to work with both v2 and v3.

Fixes zendframework/zend-navigation#19 and zenframework/zend-view#45

### DIFF
--- a/src/Helper/Navigation/PluginManager.php
+++ b/src/Helper/Navigation/PluginManager.php
@@ -49,6 +49,13 @@ class PluginManager extends HelperPluginManager
         Links::class       => InvokableFactory::class,
         Menu::class        => InvokableFactory::class,
         Sitemap::class     => InvokableFactory::class,
+
+        // v2 canonical FQCNs
+
+        'zendviewhelpernavigationbreadcrumbs' => InvokableFactory::class,
+        'zendviewhelpernavigationlinks'       => InvokableFactory::class,
+        'zendviewhelpernavigationmenu'        => InvokableFactory::class,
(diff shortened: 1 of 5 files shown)


weierophinney:
Most recent changes also address the alias resolution reported in #46.

weierophinney:
In a private email, @webimpress has indicated that the initializers registered by `HelperPluginManager` are not receiving a container instance; I'm going to investigate that before merging.

weierophinney:
Okay, what I've discovered is that `Zend\View\Helper\Navigation` lazy-loads `Zend\View\Helper\Navigation\PluginManager` by default. This means that it doesn't have a parent service manager instance, which in turn means that the initializers it registers due to inheriting from `Zend\View\HelperPluginManager` do not have a parent locator instance, and thus cannot be injected with things like the translator.

Under v2, this worked because of this line:

- https://github.com/zendframework/zend-view/blob/f1b682652045b126e0eb0852a1c9fb76b21dfcdd/src/HelperPluginManager.php#L153-L155

which essentially made it so that the navigation plugins, even if translator-aware, are *never* injected with a translator.

I'll re-instate that behavior for now, and later, when preparing zend-navigation for forwards-compatibility, I'll set it up to create a navigation plugin manager that's injected with the parent locator of the helper plugin manager.

froschdesign:
Looks good!

gianarb:
Good :+1: 2.6.2 works fine for me.. Thanks

PR #54, opened 2016-03-17
Updates to use lowest/locked/latest strategy for continuous integration

weierophinney:
- Adds `composer.lock` to the repository
- Updates `.travis.yml` to run, for each supported PHP version:
  - against the lowest supported versions (`composer update --prefer-lowest`)
  - against the versions in the lockfile (`composer install`)
  - against the latest versions (`composer update`)

In order for the `latest` target to work, PHPUnit must be a minimum 4.5 version.


weierophinney:
Ping @stefanotorresi - you don't need to do this one. :)

(Did this in an attempt to verify a reported issue.)

PR #60, opened 2016-05-06
Allow view helpers to be simple callables

DASPRiD:
Currently, the user is forced to implement a `setView()` and `getView()` method in his view helpers, even though he's mostly likely not going to use them. This PR solves this by simply allowing any callable as a view helper. Extending classes can still specify a specific interface or class to check for, but the primary helper plugin manager now accepts any callable. This **should** be no BC break, but @weierophinney probably knows best.

### DIFF
--- a/src/HelperPluginManager.php
+++ b/src/HelperPluginManager.php
@@ -137,7 +137,7 @@ class HelperPluginManager extends AbstractPluginManager
         'ViewModel'           => Helper\ViewModel::class,
     ];
 
-    protected $instanceOf = Helper\HelperInterface::class;
+    protected $instanceOf = 'callable';
 
     /**
      * Default factories
@@ -289,6 +289,10 @@ class HelperPluginManager extends AbstractPluginManager
             ? $second
             : $first;
 
+        if (! $helper instanceof Helper\HelperInterface) {
+            return;
+        }
+
         $renderer = $this->getRenderer();
         if (null === $renderer) {
             return;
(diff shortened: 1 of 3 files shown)


src/HelperPluginManager.php
```diff
@@ -400,7 +404,17 @@ public function injectEventManager($first, $second)
...
-        if (!$instance instanceof $this->instanceOf) {
+        if ('callable' === $this->instanceOf) {
+            if (is_callable($instance)) {
+                return;
+            }
+
+            $instanceOf = Helper\HelperInterface::class;
+        } else {
+            $instanceOf = $this->instanceOf;
+        }
```
weierophinney: We can make this simpler by removing the `$instanceOf` property entirely. The logic then becomes:

```php
public function validate($instance)
{
    if (is_callable($instance) || $instance instanceof Helper\HelperInterface) {
       return;
    }
    // throw exception
}
```

I can do that on merge.

test/HelperPluginManagerCompatibilityTest.php
```diff
@@ -58,7 +57,7 @@ protected function getV2InvalidPluginException()
 
     protected function getInstanceOf()
     {
-        return HelperInterface::class;
+        return 'callable';
```
weierophinney: Of course, removing the `$instanceOf` property might affect this test. I'll try it out and see.

weierophinney:
Merged to develop for release with 2.7.0.

(This *loosens* the types allowed, making it BC.)

PR #63, opened 2016-05-26
remove composer.lock

basz:
https://gist.github.com/basz/2575cdb5e6a0dd482ed19b2e4be02696


thomasvargiu:
I think it's necessary for testing. If you take a look at `.travis.yml` file there is a matrix to:

- tests with locked dependencies (last releases during development using `composer.lock`)
- tests with lowest (with `composer update --prefer-lowest` to ensure that lowest dependencies are compatible)
- test with latest (with `composer update` to ensure that latest dependency releases are compatible)

I think it's a good choice, so `composer.lock` is required.

basz:
https://github.com/zendframework/zend-inputfilter/pull/107#issuecomment-221933966

PR #64, opened 2016-06-12
Asset helper

webimpress:
Asset view helper. It required resource map (array) configuration. Can be used also with JSON file (eg. generated by gulp/grunt - rev-manifest.json) as described in docs.

I am not 100% happy with tests because of confusing compatibility with zend-servicemanager v2 and v3.

Any suggestions would be appreciated.

### DIFF
--- a/doc/book/helpers/asset.md
+++ b/doc/book/helpers/asset.md
@@ -37,15 +37,15 @@ The first argument of the `asset` helper is the regular asset name,
 which will be replaced by versioned asset name defined in `resource_map`
 of the configuration.
 
-### Note
-
-When `asset` key is defined but `resource_map` is not provided or is not
-an array exception `Zend\View\Exception\RuntimeException` will be
+> ### Note
+>
+> When `asset` key is defined but `resource_map` is not provided or is not
+> an array exception `Zend\View\Exception\RuntimeException` will be
 thrown.
-
-When you call `asset` helper with parameter which is not defined on your
-`resource_map` exception `Zend\View\Exception\InvalidArgumentException`
-will be thrown.
+>
+> When you call `asset` helper with parameter which is not defined on your
+> `resource_map` exception `Zend\View\Exception\InvalidArgumentException`
+> will be thrown.
 
 ## Resource map in JSON file
 
--- a/doc/book/helpers/asset.md
+++ b/doc/book/helpers/asset.md
@@ -1,0 +1,74 @@
+# Asset
+
+The `Asset` helper is used to translate asset names.
+It could be used to prevent browser caching for assets.
+
+## Configuration and Basic Usage
+
+`Zend\View\Helper\Service\AssetFactory` checks the application
+configuration, making it possible to set up the resource map through
+your `module.config.php`. The next example will set up `Asset` helper:
+
+```php
+'view_helper_config' => [
+    'asset' => [
+        'resource_map' => [
+            'css/style.css' => 'css/style-3a97ff4ee3.css',
+            'js/vendor.js' => 'js/vendor-a507086eba.js',
+        ],
+    ],
+],
+```
+
+Then in your view you can use:
+
+```php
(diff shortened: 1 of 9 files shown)


weierophinney:
This looks like a nice solution in terms of being light-weight, and integrating with established front-end workflows.

My one comment: I wonder if this should integrate with `basePath()`? As an example, once the path is resolved from the resource map, should that path then be passed to `basePath()` in order to ensure it is resolved correctly? My inclination is yes, but the question is if the helper should do that internally, or if we should document that the workflow should be:

```php
<?= $this->basePath($this->asset('css/style.css')) ?>
```

In particular, the use of relative URLs in the resource map may be problematic when on a path that represents more than one level deep in the tree (e.g., '/foo/bar/baz' would make 'css/style.css' resolve to '/foo/bar/css/style.css' on most browsers). If the asset paths are expected to be relative to the application root, we need to either enforce that within the helper, or document that the developer is responsible for this.

Thoughts?

webimpress:
@weierophinney good point. I thought about it, and I think developer should be responsible to ensure he has a good paths. We can add into documentation example with `basePath` helper, everything depends on resource map array - if there are relative or absolute paths.

I think better if we allow decide developers how they want use it. Probably it will be used mostly with plugins `headLink`, `headScript`, ... I can imagine also that maybe somebody will use it as following:

```
'view_helper_config' => [
    'asset' => [
        'resource_map' => [
            'jquery' => 'https://code.jquery.com/jquery-3.0.0.min.js',
        ],
    ],
],
```

and in view:

```<?= $this->headScript()->appendFile($this->asset('jquery')) ?>```

It's hard to say how it will be used, but I think without integration with `basePath` we have a greater capacity of use.

BTW. How you'd like to use one plugin inside another? Pass it via factory?

File: doc/book/helpers/asset.md [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+
+```html
+css/style-3a97ff4ee3.css
+```
+
+The first argument of the `asset` helper is the regular asset name,
+which will be replaced by versioned asset name defined in `resource_map`
+of the configuration.
+
+### Note
```
froschdesign: Please put the note in a blockqoute.

```
> ### Note
> ...

File: src/Helper/Asset.php [hunk 2]
```diff
@@ -0,0 +1,48 @@
...
+    public function __invoke($asset)
+    {
+        if (!array_key_exists($asset, $this->resourceMap)) {
+            throw new Exception\InvalidArgumentException('Asset is not defined.');
+        }
+
+        return $this->resourceMap[$asset];
+    }
+
+    public function setResourceMap($resourceMap)
```
froschdesign: Add a full DocBlock. (description, parameter, return value)

src/Helper/Asset.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $this->resourceMap = $resourceMap;
+
+        return $this;
+    }
+
+    public function getResourceMap()
```
froschdesign: Same here.

src/Helper/Asset.php
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+namespace Zend\View\Helper;
+
+use Zend\View\Exception;
+
+/**
+ * View helper plugin to fetch asset from resource map.
+ */
+class Asset extends AbstractHelper
```
froschdesign: Please add the helper in the DocBlock of [`PhpRenderer`](https://github.com/zendframework/zend-view/blob/master/src/Renderer/PhpRenderer.php#L34) for code completion in IDEs.

webimpress: Done 3e08baf

webimpress:
@froschdesign I've made all requested changes and added also a few fixes

weierophinney:
Thanks, @webimpress

PR #65, opened 2016-06-15
Added template map generator as bin script

RalfEggert:
Added template map generator bin script and added it as well to the composer.json file for easy installation

PR #67, opened 2016-06-20
Updated templatemap_generator to remove dep on zend-console

weierophinney:
This patch builds on #65, incorporating the feedback I provided via [a comment on the original ZendSkeletonApplication issue](https://github.com/zendframework/ZendSkeletonApplication/pull/340#discussion_r67228492) that began the change.

Essentially, I want to avoid adding zend-console as a dependency of zend-view.

What this patch does is remove the usage of `Zend\Console\Getopt`, and simplifies the usage to a single, narrow use case: `cd module/ModuleName/config && ../../../vendor/bin/templatemap_generator.php ../view ../view/**/*.phtml > template_map.config.php`. This command will create the file `module/ModuleName/config/template_map.config.php`, which will map templates to `.phtml` view scripts located in `module/ModuleName/view/`. Users would then define the template map entry in `module.config.php` as follows:

```php
'template_map' => include __DIR__ . '/template_map.config.php',
```

This allows regeneration of the template map on-demand using the same command, without requiring parsing and updating the main module configuration file.

I believe this is the 80% use case, and developers can build out their own solutions for anything needing more customization.

Full usage is:

```text
Generate template maps.

Usage:

templatemap_generator.php [-h|--help] templatepath <files...>

--help|-h                    Print this usage message.
templatepath                 Path to templates relative to current working
                             path; used to identify what to strip from
                             template names. Must be a directory.
<files...>                   List of files to include in the template
                             map, relative to the current working path.

The script assumes that paths included in the template map are relative
to the current working directory.

The script will output a PHP script that will return the template map
on successful completion. You may save this to a file using standard
piping operators; use ">" to write to/ovewrite a file, ">>" to append
to a file (which may have unexpected and/or intended results; you will
need to edit the file after generation to ensure it contains valid
PHP).

If only the templatepath argument is provided, the script will look for
all .phtml files under that directory, creating a map for you.

If you want to specify a specific list of files -- for instance, if you
are using an extension other than .phtml -- we recommend one of the
following constructs:

For any shell, you can pipe the results of `find`:

    $(find ../view -name '*.phtml')

For zsh, or bash where you have enabled globstar (`shopt  -s globstar` in
either your bash profile or from within your terminal):

    ../view/**/*.phtml

We recommend you then include the generated file within your module
configuration:

  'template_map' => include __DIR__ . '/template_map.config.php',

Examples:

  # Using only a templatepath argument, which will match any .phtml
  # files found under the provided path:
  $ cd module/Application/config/
  $ ../../../vendor/bin/templatemap_generator.php ../view > template_map.config.php

  # Create a template_map.config.php file in the Application module's
  # config directory, relative to the view directory, and only containing
  # .html.php files; overwrite any existing files:
  $ cd module/Application/config/
  $ ../../../vendor/bin/templatemap_generator.php ../view ../view/**/*.html.php > template_map.config.php

  # OR using find:
  $ ../../../vendor/bin/templatemap_generator.php \
  > ../view \
  > $(find ../view -name '*.html.php') > template_map.config.phpp
```

### DIFF
--- a/bin/templatemap_generator.php
+++ b/bin/templatemap_generator.php
@@ -6,21 +6,6 @@
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
 
-// Setup/verify autoloading
-if (file_exists(__DIR__ . '/../vendor/autoload.php')) {
-    // Local install
-    require __DIR__ . '/../vendor/autoload.php';
-} elseif (file_exists(getcwd() . '/vendor/autoload.php')) {
-    // Root project is current working directory
-    require getcwd() . '/vendor/autoload.php';
-} elseif (file_exists(__DIR__ . '/../../../autoload.php')) {
-    // Relative to composer install
-    require __DIR__ . '/../../../autoload.php';
-} else {
-    fwrite(STDERR, "Unable to setup autoloading; aborting\n");
-    exit(2);
-}
-
 $help = <<< EOH
 Generate template maps.
 
--- a/bin/templatemap_generator.php
+++ b/bin/templatemap_generator.php
@@ -47,27 +47,27 @@ EOH;
 // Called without arguments
 if ($argc < 2) {
     fwrite(STDERR, 'No arguments provided.' . PHP_EOL . PHP_EOL);
-    fwrite(STDERR, $help);
+    fwrite(STDERR, $help . PHP_EOL);
     exit(2);
 }
 
 // Requested help
 if (in_array($argv[1], ['-h', '--help'], true)) {
-    echo $help, "\n";
+    echo $help, PHP_EOL;
     exit(0);
 }
 
 // Invalid path argument
 if (! is_dir($argv[1])) {
     fwrite(STDERR, 'templatepath argument is not a directory.' . PHP_EOL . PHP_EOL);
-    fwrite(STDERR, $help);
+    fwrite(STDERR, $help . PHP_EOL);
     exit(2);
 }
 
 // Not enough arguments
 if ($argc < 3) {
     fwrite(STDERR, 'No files specified.' . PHP_EOL . PHP_EOL);
-    fwrite(STDERR, $help);
+    fwrite(STDERR, $help . PHP_EOL);
     exit(2);
 }
 
--- a/bin/templatemap_generator.php
+++ b/bin/templatemap_generator.php
@@ -30,6 +30,11 @@ to a file (which may have unexpected and/or intended results; you will
 need to edit the file after generation to ensure it contains valid
 PHP).
 
+We recommend you then include the generated file within your module
+configuration:
+
+  'template_map' => include __DIR__ . '/template_map.config.php',
+
 If only the templatepath argument is provided, the script will look for
 all .phtml files under that directory, creating a map for you.
 
@@ -45,11 +50,6 @@ For zsh, or bash where you have enabled globstar (`shopt  -s globstar` in
 either your bash profile or from within your terminal):
 
     ../view/**/*.phtml
-
-We recommend you then include the generated file within your module
-configuration:
-
-  'template_map' => include __DIR__ . '/template_map.config.php',
 
 Examples:
 
--- a/bin/templatemap_generator.php
+++ b/bin/templatemap_generator.php
@@ -30,6 +30,17 @@ to a file (which may have unexpected and/or intended results; you will
 need to edit the file after generation to ensure it contains valid
 PHP).
 
+To provide a list of files, we recommend using one of the following.
+
+For any shell, you can pipe the results of `find`:
+
+    $(find ../view -name '*.phtml')
+
+For zsh, or bash where you have enabled globstar (`shopt  -s globstar` in
+either your bash profile or from within your terminal):
+
+    ../view/**/*.phtml
+
 We recommend you then include the generated file within your module
 configuration:
 
@@ -42,6 +53,11 @@ Examples:
   # .phtml files; overwrite any existing files:
   $ cd module/Application/config/
   $ ../../../vendor/bin/templatemap_generator.php ../view ../view/**/*.phtml > template_map.config.php
+
+  # OR using find:
+  $ ../../../vendor/bin/templatemap_generator.php \
+  > ../view \
+  > $(find ../view -name '*.phtml') > template_map.config.php
 EOH;
 
 // Called without arguments
--- a/bin/templatemap_generator.php
+++ b/bin/templatemap_generator.php
@@ -1,30 +1,9 @@
 #!/usr/bin/env php
 <?php
 /**
- * Zend Framework (http://framework.zend.com/)
- *
  * @link      http://github.com/zendframework/zend-view for the canonical source repository
  * @copyright Copyright (c) 2005-2016 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   http://framework.zend.com/license/new-bsd New BSD License
- */
-
-use Zend\Console;
-
-/**
- * Generate template maps.
- *
- * Usage:
- * --help|-h                    Get usage message
- * --library|-l [ <string> ]    Library to parse; if none provided, assumes
- *                              current directory
- * --view|-v [ <string> ]       View path to parse; if none provided, assumes
- *                              view as template directory
- * --extensions|-e [ <string> ] List of accepted file extensions (regex alternation
- *                              without parenthesis); default: *
- * --output|-o [ <string> ]     Where to write map file; if not provided,
- *                              assumes "template_map.php" in library directory
- * --append|-a                  Append to map file if it exists
- * --overwrite|-w               Whether or not to overwrite existing map file
  */
 
 // Setup/verify autoloading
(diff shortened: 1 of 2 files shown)


weierophinney:
@RalfEggert pinging for your review, as you started this whole thing. :smile:

Does the use case as demonstrated above capture what you would most likely need?

webimpress:
Sorry, guys, do we definitely need it? What is the advantage of this script over the @Ocramius library
https://github.com/Ocramius/OcraCachedViewResolver

Let me quote @Ocramius:
> This, in my opinion, is micro-optimization and complication.
https://github.com/zendframework/ZendSkeletonApplication/pull/340#discussion_r67231459

@Ocramius Could you just update your library to support also PHP 5.6? Is there any problem with it?

weierophinney:
@webimpress As you've already noted, OcraCachedViewResolver currently requires PHP 7 (though you can install the v3 versions if you need 5.5 or 5.6 support); additionally, it requires usage of zend-cache, which adds a number of extra dependencies beyond what we require by default in zend-mvc, and which then requires setting up a cache storage backend. The approach of `templatemap_generator.php` is deliberately simple and inflexible, to cover the general use case of wanting a static map that is stored as part of your application/module repository. The two approaches are not mutually exclusive.

webimpress:
@weierophinney thanks for explanations. It makes sense.

It would be nice to have both approaches described in documentation so developers could decide what they want use. Obviously `OcraCachedViewResolver` has bigger functionality, but as you said also bigger requirements.
Anyway, I hope @Ocramius could add support for PHP 5.6 in v4 (because v3 does not support ZF3 modules). For now then his library can't be used with ZF3 on PHP5.6... :(

bin/templatemap_generator.php
```diff
@@ -0,0 +1,111 @@
...
+
+    $template = (0 === strpos($template, $basePath))
+        ? substr($template, strlen($basePath))
+        : $template;
+
+    $template = preg_match('#(?P<template>.*?)\.[a-z0-9]+$#i', $template, $matches)
+        ? $matches['template']
+        : $template;
+
+    $template = str_replace('\\', '/', $template);
```
webimpress: Isn't that line redundant (because of line 86)?

bin/templatemap_generator.php
```diff
@@ -0,0 +1,111 @@
...
+    $template = preg_replace('#^\.*/#', '', $template);
+    $file     = sprintf('__DIR__ . \'%s\'', $file);
```
webimpress: I think we need one more directory separator after `__DIR__` and before file name. This is the script results on ZendSkeletonApplication (develop, v3):

```
return [
    'view/error/index' => __DIR__ . '../view/error/index.phtml',
    'view/layout/layout' => __DIR__ . '../view/layout/layout.phtml',
];
```

Hm.... So I think the result is wrong at all... I think we should get something like that:
```
<?php
return [
    'application/index/index' => __DIR__ . '/../view/application/index/index.phtml',
    'error/404' => __DIR__ . '/../view/error/404.phtml',
    'error/index' => __DIR__ . '/../view/error/index.phtml',
    'layout/layout' => __DIR__ . '/../view/layout/layout.phtml',
];
```

webimpress: Yeah, definitely it's wrong now, this is result from previous script (PR #65):
`$ vendor/bin/templatemap_generator.php --view=module/Application/view/ --output=module/Application/config/template_map.php`:

```
<?php
// Generated by Zend\View's ./bin/templatemap_generator.php
return array(
    'application/index/index' => __DIR__ . '/../../../module/Application/view/application/index/index.phtml',
    'error/404'               => __DIR__ . '/../../../module/Application/view/error/404.phtml',
    'error/index'             => __DIR__ . '/../../../module/Application/view/error/index.phtml',
    'layout/layout'           => __DIR__ . '/../../../module/Application/view/layout/layout.phtml',
);
```

Maybe it's not perfect as well but it works as I would expect.

bin/templatemap_generator.php
```diff
@@ -0,0 +1,111 @@
...
+    // Local install
+    require __DIR__ . '/../vendor/autoload.php';
+} elseif (file_exists(getcwd() . '/vendor/autoload.php')) {
+    // Root project is current working directory
+    require getcwd() . '/vendor/autoload.php';
+} elseif (file_exists(__DIR__ . '/../../../autoload.php')) {
+    // Relative to composer install
+    require __DIR__ . '/../../../autoload.php';
+} else {
+    fwrite(STDERR, "Unable to setup autoloading; aborting\n");
```
webimpress: Can we be consistent and have also `echo` here?

bin/templatemap_generator.php
```diff
@@ -0,0 +1,111 @@
...
+    exit(2);
+}
```
webimpress: Do we need this whole autoloader part? Not sure about it.

bin/templatemap_generator.php
```diff
@@ -0,0 +1,111 @@
...
+  # Create a template_map.config.php file in the Application module's
+  # config directory, relative to the view directory, and only containing
+  # .phtml files; overwrite any existing files:
+  $ cd module/Application/config/
+  $ ../../../vendor/bin/templatemap_generator.php ../view/**/*.phtml > template_map.config.php
+EOH;
+
+// Called without arguments
+if ($argc < 2) {
+    echo $help;
```
webimpress: Don't we need `"\n"` / `PHP_EOL` here?

webimpress:
Sorry again... Maybe idea to have the script completely independent of other libraries is good, but as we can see now the script is not working properly, however it's much simpler than previous version. To get working, as expected, we need make some changes and it will complicate script again. What's more the script have to be maintain (yeah, you would say that is simply script, and once written will work forever!). There is also no any tests for it so we don't even know what is the expected result (we can only assume or test it manually).

I don't like the procedural form of the script, which I think it's not on the ZF3 level at all. Maybe it should be rethink and done again properly (?)

bin/templatemap_generator.php
```diff
@@ -0,0 +1,111 @@
...
+
+  'template_map' => include __DIR__ . '/template_map.config.php',
+
+Examples:
+
+  # Create a template_map.config.php file in the Application module's
+  # config directory, relative to the view directory, and only containing
+  # .phtml files; overwrite any existing files:
+  $ cd module/Application/config/
+  $ ../../../vendor/bin/templatemap_generator.php ../view/**/*.phtml > template_map.config.php
```
webimpress: As it is said it gives results:
```
return [
    'view/error/index' => __DIR__ . '../view/error/index.phtml',
    'view/layout/layout' => __DIR__ . '../view/layout/layout.phtml',
];
```

As description above says probably it should be:
`$ ../../../vendor/bin/templatemap_generator.php ../view ../view/**/*.phtml > template_map.config.php`
and then the result is:
```
<?php
return [
    'error/404' => __DIR__ . '../view/error/404.phtml',
    'error/index' => __DIR__ . '../view/error/index.phtml',
    'layout/layout' => __DIR__ . '../view/layout/layout.phtml',
];
```

So there are still missing:
```
    'application/index/index' => __DIR__ . '/../view/application/index/index.phtml',
```

bin/templatemap_generator.php
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Requested help
+if (in_array($argv[1], ['-h', '--help'], true)) {
+    echo $help, "\n";
+    exit(0);
+}
+
+// Not enough arguments
+if ($argc < 3) {
```
webimpress: Probably it is not enough, because invalid execution - as given above generate wrong result:
```
$ cd module/Application/config/
$ ../../../vendor/bin/templatemap_generator.php ../view/**/*.phtml > template_map.config.php
```
Maybe then there should be check also if `is_dir($argv[2])` ?

bin/templatemap_generator.php
weierophinney: Yep, thanks for the catch!

bin/templatemap_generator.php
weierophinney: I'm not seeing the same behavior as you. I tried just now using the same arguments (though with a path to my local working checkout of zend-view for the script), and received the following:

```php
<?php
return [
    'application/index/index' => __DIR__ . '../view/application/index/index.phtml',
    'error/404' => __DIR__ . '../view/error/404.phtml',
    'error/index' => __DIR__ . '../view/error/index.phtml',
    'layout/layout' => __DIR__ . '../view/layout/layout.phtml',
    'partial/breadcrumb' => __DIR__ . '../view/partial/breadcrumb.phtml',
    'partial/paginator' => __DIR__ . '../view/partial/paginator.phtml',
];
```

(I have a couple more templates in mine due to having done some extra tutorials in my skeleton checkout.)

The only bit wrong with it is the missing `/` at the start of the quoted path, which I'll add now; otherwise, I'm getting all entries (and not missing the `application/index/index` as you observe).

bin/templatemap_generator.php
weierophinney: No, because that message needs to go to STDERR (vs STDOUT). I'll update the two help cases that exit with status 2 to write to STDERR as well, though.

bin/templatemap_generator.php
weierophinney: It's necessary, as the autoloading path differs based on whether it's run within the checkout, from a symlink, or based on the bin-dir where composer injects it. (I've worked with this a ton; if I could omit a case, I would.)

bin/templatemap_generator.php
weierophinney: There's one at the end of the `$help` string already. :smile:

bin/templatemap_generator.php
weierophinney: I fixed the example for the help message locally. I still cannot reproduce your results; I'm getting the `application/index/index` entry.

weierophinney:
I've incorporated your feedback; thanks, @webimpress !

bin/templatemap_generator.php
webimpress: Hm... so what classes in that script we load using composer autoloader? Once you removed all dependencies, I think it's useless to require `autoload.php` file.

weierophinney: Oh, that's what you were getting at! Good point; removing!

webimpress:
@weierophinney thanks for fixes, but still I have the same results:
```
$ ../../../vendor/bin/templatemap_generator.php ../view ../view/**/*.phtml
<?php
return [
    'error/404' => __DIR__ . '/../view/error/404.phtml',
    'error/index' => __DIR__ . '/../view/error/index.phtml',
    'layout/layout' => __DIR__ . '/../view/layout/layout.phtml',
];
```

Maybe it will help:
```
$ echo ../view/***/*.phtml
../view/error/404.phtml ../view/error/index.phtml ../view/layout/layout.phtml

$ echo ../view/**/**/*.phtml
../view/application/index/index.phtml
```

Do you think it could be system dependent? I'm using OS X but the same results I've got on Windows.

webimpress:
@weierophinney please update also docs above in PR description, these still is wrong:
`$ ../../../vendor/bin/templatemap_generator.php ../view/**/*.phtml > template_map.config.php`

Do we definitely need the 2nd and 3rd parameter? Maybe will be enough if we pass only path to the view directory and generate templates map of all files in this directory (and subdirectories)? Now it looks a bit redundant. I know it gives us bigger capacities of use, but I wonder how important it is. Previous script has much more parameters, and you've already removed some.

Thoughts?

weierophinney:
> Do you think it could be system dependent? I'm using OS X but the same results I've got on Windows.

Yes - just got everything setup on my Windows VM to test, and can confirm. I suspect it's a difference in globbing support; I've been testing on zsh, but I know that in Windows, where I run Cygwin, I've got bash, which is also the default shell on Mac. If so, it's just a matter of documentation.

weierophinney:
> please update also docs above in PR description

Done.

webimpress:
> If so, it's just a matter of documentation.

Yeah, but if we would like generate templates map for whole directory (with all subdirectories) it will be complicated, so maybe better as I've suggested above to pass only path to the `view` directory. What do you think?

weierophinney:
> Do we definitely need the 2nd and 3rd parameter? Maybe will be enough if we pass only path to the view directory and generate templates map of all files in this directory (and subdirectories)? Now it looks a bit redundant. I know it gives us bigger capacities of use, but I wonder how important it is. Previous script has much more parameters, and you've already removed some.

What I tried to do for this iteration of the script is follow the Unix philosophy, along with poka-yoke principles. The idea is to keep usage to a specific pattern; if you go outside that, you need to use another tool or manually create the map.

Part of this was due to wanting to remove dependencies; I see no reason why zend-view should require zend-console, particularly if it's only for this script. The other part is because the previous version had a ton of execution paths, and I'm positive not all of them were tested. Reducing the use cases helps keep it focused and ensure it works correctly.

> > If so, it's just a matter of documentation.
> Yeah, but if we would like generate templates map for whole directory (with all subdirectories) it will be complicated, so maybe better as I've suggested above to pass only path to the view directory. What do you think?

I discovered two ways to make this work in bash.

First, and most portable across all shells, replace `../view/**/*.phtml` with `$(find ../view -name '*.phtml')`. This is a bit unwieldy.

Second, Bash 4 introduced a `globstar` shell option that, for some reason, is not enabled by default. You can enable it with:

```bash
shopt -s globstar
```

This can be done in either your bash profile, or on the command line. Once enabled, the `**/*.phtml` version works correctly.

I'll update the help doc to indicate these various forms.

webimpress:
@weierophinney one more suggestion: maybe the 3rd parameter shouldn't be mandatory, so it will be possible generate templates map for all files in provided view `directory`? Wouldn't it be nicer for normal use?
Your both ways are good, works, just tested, but probably not obvious and hard to remember. I know, they are described in documentation and script help, but wouldn't be nicer just call:
`$ ../../../vendor/bin/templatemap_generator.php ../view` ? 😄

weierophinney:
> one more suggestion: maybe the 3rd parameter shouldn't be mandatory, so it will be possible generate templates map for all files in provided view directory?

This would require adding a flag to indicate what view script extension to use. Using globbing means we don't need to add that flag, or assume `.phtml` for that use case.

Let me think about it for a bit; I definitely see your point, and we could *assume* that if no files are given, `.phtml` is assumed for the suffix; users would specify files if they want only those with a specific suffix.

weierophinney:
@webimpress — just pushed a commit that allows the following:

```bash
$ templatemap_generator.php ../view > template_map.config.php
```

This uses a leaves-only recursive directory search to find all `.phtml` files in that given path. If users want to specify an alternate extension, they can use file globbing or `find`. I'll update the help message in the issue description, and then I think this is ready. Thanks for the great feedback!

RalfEggert:
@weierophinney 

Sorry for the delay. I only find time now to look at this. Looks really great. I just notices one little issue so far. I used the template map generator for a Zend Skeleton Application project like mentioned in the docs:

```
  $ cd module/Application/config/
  $ ../../../vendor/bin/templatemap_generator.php ../view ../view/**/*.phtml > template_map.config.php
```

That created this file:

```
<?php
return [
    'error/404' => __DIR__ . '/../view/error/404.phtml',
    'error/index' => __DIR__ . '/../view/error/index.phtml',
    'layout/layout' => __DIR__ . '/../view/layout/layout.phtml',
];
```

When I use this

```
  $ cd module/Application/config/
  $ ../../../vendor/bin/templatemap_generator.php ../view > template_map.config.php
```

I get that file:

```
<?php
return [
    'application/index/index' => __DIR__ . '/../view/application/index/index.phtml',
    'error/index' => __DIR__ . '/../view/error/index.phtml',
    'error/404' => __DIR__ . '/../view/error/404.phtml',
    'layout/layout' => __DIR__ . '/../view/layout/layout.phtml',
];
```

I guess this is because of the extra directory level for the `application/index/index` file?

webimpress:
@RalfEggert please read PR description or help of the script - @weierophinney described two approaches to resolve that issue

RalfEggert:
@webimpress Well, to be honest, I only looked in the examples of the inline docs of that file. And that caused the missunderstanding. Maybe needs to be updated?

webimpress:
@RalfEggert yeah, I had the same feelings, so I've asked @weierophinney to change files list to be not mandatory parameter... Please have a look on @weierophinney comment: https://github.com/zendframework/zend-view/pull/67#issuecomment-227548858

RalfEggert:
:+1:

PR #68, opened 2016-06-21
Travis yml coverage fix

Koopzington:
--no-update flag needs to be removed in order to make installation of satooshi/php-coveralls work with an existing composer.lock.


Koopzington:
The execution of coveralls had to be moved to the latest-env because it would otherwise result in composer doing a full update instead of using the versions mentioned in the composer.lock

weierophinney:
Thanks, @Koopzington

PR #78, opened 2016-06-29
V3 Compatibility

Rockstar04:


Koopzington:
You'll need to add the [Flashmessenger plugin](https://github.com/zendframework/zend-mvc-plugin-flashmessenger) to the composer.json and update the composer.lock in order to let that build pass.

Rockstar04:
Then that may need a larger talk about removing this from zend-view all together and moving it to the flash messenger package.

weierophinney:
What we need to do for the interim is as follows:

- Add the flash messenger plugin to the dev requirements
- Modify the helper to remove the typehint in `setPluginFlashMessenger()`, and have it do checking for either the v2 version or the standalone package version, raising an InvalidArgumentException if neither.
- We can add the helper into the standalone package, and it would then override configuration for the helper. If you'd like to take that on, please be my guest; I've [opened an issue to track it](zendframework/zend-mvc-plugin-flashmessenger#3).
- When we're ready to do a v3 of this component, we can remove the helper.

I'll take your PR as a starting point, and do the items we can do now to provide forwards compatibility.

Rockstar04:
Awesome, thanks Matthew!

PR #79, opened 2016-06-30
Provide forward compatibility with v3 where functionality is broken

weierophinney:
Updates zend-view to be forwards compatible with zend-mvc v3.

## UrlHelper

The UrlHelper tests were previously using typehints against the zend-mvc v2 router in some locations. This patch provides updates to the test to detect which router version is present, as well as which route classes, so it can test correctly across both router versions.

## Navigation helpers

The navigation helper tests were:

- hardcoding against the v2 RouteMatch
- not taking into account that route services move to the zend-router package when zend-mvc v3 is present.

They have been updated to resolve this, and now pass against both versions.

## FlashMessenger

Updates the FlashMessenger to be forwards compatible with zendframework/zend-mvc-plugin-flashmessenger, while retaining compatibility with zend-mvc 2.7. This required:

- Removing references to class constants defined in the FlashMessenger MVC plugin; this means temporarily hardcoding the values in the helper, and resulted in a TODO item to re-introduce them for version 3.
- Removing typehints for the FlashMessenger MVC plugin; when required, we now test against both the v2 and v3 class name to validate.
- Deprecating the helper for version 3.0.

This PR supercedes #78.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,11 @@ cache:
 env:
   global:
     - COMPOSER_ARGS="--no-interaction --ignore-platform-reqs"
-    - SITE_URL: https://zendframework.github.io/zend-view
-    - GH_USER_NAME: "Matthew Weier O'Phinney"
-    - GH_USER_EMAIL: matthew@weierophinney.net
-    - GH_REF: github.com/zendframework/zend-view.git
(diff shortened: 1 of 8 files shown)


weierophinney:
WE HAVE GREEN!!!!

PR #83, opened 2016-08-09
Typo viewHelperManager > ViewHelperManager

Ponsjuh:
this results in actually finding the ViewHelperManager using lowercase will result in 
Uncaught Zend\ServiceManager\Exception\ServiceNotFoundException: Unable to resolve service "viewHelperManager" to a factory;

### DIFF
--- a/doc/book/helpers/head-title.md
+++ b/doc/book/helpers/head-title.md
@@ -56,7 +56,7 @@ class Module
         $siteName   = 'Zend Framework';
 
         // Getting the view helper manager from the application service manager
-        $viewHelperManager = $e->getApplication()->getServiceManager()->get('viewHelperManager');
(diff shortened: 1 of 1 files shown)


adamlundrigan:
@weierophinney should the documentation just be updated to fix the typo, or should `zend-view` also provide an alias to fix these common typos silently (like [`zend-mvc-plugin-flashmessenger`](https://github.com/zendframework/zend-mvc-plugin-flashmessenger/blob/master/src/Module.php#L26-L28) does)?

weierophinney:
@adamlundrigan In this particular case, the service is registered via `Zend\Mvc\Service\ServiceListenerFactory`; as such, we'd need to add the aliases there. If you would like to do that, go for it; I think updating the docs makes sense, though, as we typically only want a single canonical name for the various top-level application services.

weierophinney:
Thanks, @Ponsjuh

PR #84, opened 2016-08-11
DOCS: Dont mention the inspection of the Accept HTTP Header

MatthiasKuehneEllerhold:
The View Strategies are mentioning (in the code and in the docs) that they are inspecting the "Accept" HTTP header and determine if they are going to inject their renderer and response.

But in fact, they dont do that. They just test if the View Model is an instanceof their ViewModel (JsonModel and FeedModel). I trusted the documentation on that and spent a lot of time wondering why it doesnt work... :(

I've removed every (wrong) mentioning of the inspection of the "Accept" header from the php-docs and the md docs. I've left in the (correct) sections in the quick-start on how to inspect the "Accept" header and set the correct strategy accordingly.

There should be no changes in behavior at all because only documentation was changed.

### DIFF
--- a/doc/book/quick-start.md
+++ b/doc/book/quick-start.md
@@ -583,16 +583,12 @@ within your application.
 - `Zend\View\Strategy\PhpRendererStrategy`. This strategy is a "catch-all" in
   that it will always return the `Zend\View\Renderer\PhpRenderer` and populate
   the Response body with the results of rendering.
-- `Zend\View\Strategy\JsonStrategy`. This strategy inspects the `Accept` HTTP
-  request header, if present, and determines if the client has indicated it
-  accepts an `application/json` response. If so, it will return the
+- `Zend\View\Strategy\JsonStrategy`. This strategy will return the
   `Zend\View\Renderer\JsonRenderer`, and populate the Response body with the
   JSON value returned, as well as set a `Content-Type` header with a value of
   `application/json`.
-- `Zend\View\Strategy\FeedStrategy`. This strategy inspects the `Accept` HTTP
-  header, if present, and determines if the client has indicated it accepts
-  either an `application/rss+xml` or `application/atom+xml` response. If so, it
-  will return the `Zend\View\Renderer\FeedRenderer`, setting the feed type to
+- `Zend\View\Strategy\FeedStrategy`. This strategy will return the 
+  `Zend\View\Renderer\FeedRenderer`, setting the feed type to
   either "rss" or "atom", based on what was matched. Its Response strategy will
   populate the Response body with the generated feed, as well as set a
   `Content-Type` header with the appropriate value based on feed type.
--- a/doc/book/quick-start.md
+++ b/doc/book/quick-start.md
@@ -635,7 +635,7 @@ class Module
 
 The above will register the `JsonStrategy` with the "render" event, such that it
 executes prior to the `PhpRendererStrategy`, and thus ensure that a JSON payload
-is created when the controller returns an `JsonModel`.
+is created when the controller returns a `JsonModel`.
 
 You could also use the module configuration to add the strategies:
 ```php
@@ -651,8 +651,9 @@ class Module implements \Zend\ModuleManager\Feature\ConfigProviderInterface
     public function getConfig()
     {
         return [
-            // ...
+            /* ... */
             'view_manager' => [
+                /* ... */
                 'strategies' => [
                     'ViewJsonStrategy',
                 ],
@@ -726,7 +727,7 @@ class MyController extends \Zend\Mvc\Controller\AbstractActionController
      */
     public function listAction()
     {
-        $items = /* ... get items .. .*/;
+        $items = /* ... get items ... */;
         $viewModel = new \Zend\View\Model\ViewModel();
         $viewModel->setVariable('items', $items);
         return $viewModel;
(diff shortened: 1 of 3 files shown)


adamlundrigan:
That functionality was removed from both JsonStrategy and FeedStrategy ages ago (v2.0.4, IIRC) ([PR](https://github.com/zendframework/zendframework/pull/2410)) but the documentation was not updated to match.  I believe it was replaced by the [AcceptableViewModelSelector controller plugin](http://zendframework.github.io/zend-mvc/plugins/#acceptableviewmodelselector-plugin), which requires you to define the allowable types and will juggle the ViewModel instance appropriately based on the Accept header.

MatthiasKuehneEllerhold:
@adamlundrigan Ok, one more reason to update the documentation then :)

Maybe the docs [Zend-View Quickstart](https://docs.zendframework.com/zend-view/quick-start/#creating-and-registering-alternate-rendering-and-response-strategies) should also include the link to [AcceptableViewModelSelector](http://zendframework.github.io/zend-mvc/plugins/#acceptableviewmodelselector-plugin) ?

adamlundrigan:
Yeah, that's a good idea...maybe a new blurb at the end of "Controllers and View Models" with an example of how to use the plugin.  The example at the bottom of the page (AcceptStrategy class) should also be axed as it suffers from the same security problems the Accept handling in JsonRenderer/FeedRenderer had initially.

Could you add those two changes your PR as well?

MatthiasKuehneEllerhold:
Ive looked over the whole quick start again and found another flaw:

1. "Creating and Registering Alternate Rendering and Response Strategies"
This whole section is invalid or misleading at best. Just by injecting the JsonStrategy during the render event (rather than in the config?) there is no "ensurance that a JSON payload is created when requested". 
This whole section needs to be revised to say:
* register the strategy (preferable via the config so its cacheable)
* create a JsonModel (or FeedModel) in your controller and return it
* => a JSON (or Feed) payload will be returned.

I'll try my hand on that, but english isnt my mother tongue so you'll have to except some awkward language ;)

*Offtopic:* Why do we have to enable JSON in two steps: register the strategy and explicitly return a JSON-Model. Whats the downside of always registering the two other strategies? Security, Performance, RAM? It would be a huge ease-of-use improvement if we could just return a JsonModel (or FeedModel) from a controller and it would switch the output completly.

MatthiasKuehneEllerhold:
Using the acceptableViewModelSelector didnt work (correctly) either: https://github.com/zendframework/zend-mvc/issues/194 :(

MatthiasKuehneEllerhold:
I've clarified the acceptableViewModelSelector documentation in https://github.com/zendframework/zend-mvc/pull/195
Any news on this PR?

weierophinney:
> Offtopic: Why do we have to enable JSON in two steps: register the strategy and explicitly return a JSON-Model. Whats the downside of always registering the two other strategies? Security, Performance, RAM?

Performance, flexibility, and consistency.

Performance, in that it reduces the number of view listeners required. Flexibility and consistency, in that the user is then required to register the strategy **or a substitute** at the priority required for their application; if we auto-register, there's a very real possibility that the "auto-magically" registered service might register at a priority that supercedes the user-registered service, which makes for hard-to-find error conditions. (As a real-world example, we have a separate JSON renderer in Apigility that renders HAL payloads; having ZF autoregister its own JSON strategy could lead to invalid response payloads.)

doc/book/quick-start.md
```diff
@@ -694,124 +716,46 @@ class Module
...
-use Zend\EventManager\ListenerAggregateInterface;
-use Zend\EventManager\ListenerAggregateTrait;
-use Zend\Feed\Writer\Feed;
-use Zend\View\Renderer\FeedRenderer;
-use Zend\View\Renderer\JsonRenderer;
-use Zend\View\Renderer\PhpRenderer;
+namespace Application;
 
-class AcceptStrategy implements ListenerAggregateInterface
+class MyController extends \Zend\Mvc\Controller\AbstractActionController
```
weierophinney: Import the abstract controller, and reference the imported class name.

doc/book/quick-start.md
```diff
@@ -694,124 +716,46 @@ class Module
...
-    public function attach(EventManagerInterface $events, $priority = 1)
+    /**
+     * Lists the items as HTML
+     */
+    public function listAction()
     {
-        $this->listeners[] = $events->attach('renderer', [$this, 'selectRenderer'], $priority);
-        $this->listeners[] = $events->attach('response', [$this, 'injectResponse'], $priority);
+        $items = /* ... get items ... */;
+        $viewModel = new \Zend\View\Model\ViewModel();
```
weierophinney: Import the `ViewModel` class.

doc/book/quick-start.md
```diff
@@ -694,124 +716,46 @@ class Module
...
-                $this->feedRenderer->setFeedType('atom');
-                return $this->feedRenderer;
-            }
-        }
-
-        // Nothing matched; return PhpRenderer. Technically, we should probably
-        // return an HTTP 415 Unsupported response.
-        return $this->phpRenderer;
+        $items = /* ... get items ... */;
+        $viewModel = new \Zend\View\Model\JsonModel();
```
weierophinney: Import this class as well.

doc/book/quick-start.md
```diff
@@ -694,124 +716,46 @@ class Module
...
-            // If the $result is a feed, export it
-            if ($result instanceof Feed) {
-                $result = $result->export($feedType);
-            }
-
-            $response->setContent($result);
-            return;
-        }
+        $items = /* ... get items ... */;
+        $viewModel = new \Zend\View\Model\FeedModel();
```
weierophinney: and this one.

MatthiasKuehneEllerhold:
@weierophinney Any news on this?

PR #89, opened 2016-09-14
Added "id" to allowed attributes in HeadScript view helper

### LINKED ISSUE #88: Allow scripts tag to be identified by id attribute
Using scripts view helper (`headScript` or `inlineScript`), *id* attribute is ignored when generating `scripts` tags.

```php
protected $optionalAttributes = [
    'charset',
    'crossorigin',
    'defer',
    'async',
    'language',
    'src',
];
```

Can be useful to target some scripts blocks with an html id attribute

jguittard:
Fixes #88


weierophinney:
Thanks, @jguittard

PR #98, opened 2016-11-16
Render HTML5 meta charset at first position

wandersonwhcr:
W3C Markup Validation Service currently checks if a "meta charset" is configured at first position, because a HTML5 renderer needs to know what charset to use before the content.

Makes sense.

https://validator.w3.org/
http://stackoverflow.com/questions/18007771/how-do-i-fix-error-a-charset-attribute-on-a-meta-element-found-after-the-first

### DIFF
--- a/src/Helper/HeadMeta.php
+++ b/src/Helper/HeadMeta.php
@@ -172,9 +172,18 @@ class HeadMeta extends Placeholder\Container\AbstractStandalone
         $items = [];
         $this->getContainer()->ksort();
 
+        $isHtml5 = $this->view->plugin('doctype')->isHtml5();
+
         try {
             foreach ($this as $item) {
-                $items[] = $this->itemToString($item);
+                $content = $this->itemToString($item);
+
+                if ($isHtml5 && $item->type == 'charset') {
+                    array_unshift($items, $content);
+                    continue;
+                }
+
+                $items[] = $content;
             }
         } catch (Exception\InvalidArgumentException $e) {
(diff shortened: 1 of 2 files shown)


froschdesign:
@wandersonwhcr 
> W3C Markup Validation Service currently checks if a "meta charset" is configured at first position, because a HTML5 renderer needs to know what charset to use before the content.

This description is not correct, because the HTML specification says nothing about "first position". The first 1024 bytes of the document must contain the declaration:

> The element containing the character encoding declaration must be serialized completely within the first 1024 bytes of the document.

https://www.w3.org/TR/2012/CR-html5-20121217/document-metadata.html#charset

You solution doesn't prevent the 1024 bytes problem:

```php
echo $this->headTitle(), $this->headLink(), $this->headScript(), $this->headMeta();
```

For the `headMeta` helper alone it could be a solution. And an additional hint in the documentation would be also fine.

wandersonwhcr:
Hello @froschdesign , ty for your notes about it.

BTW, I didn't mention the HTML spec, just only the W3C validator; I know about possible rendering order with `headTitle`, `headLink`, `headScript` and `headMeta`.

If you think this PR is still usefull, I can add some documentation about it.

Again, ty for your attention.

Cheers.

weierophinney:
I think it makes sense; merging for 2.8.2.

weierophinney:
Thanks, @wandersonwhcr

wandersonwhcr:
@weierophinney you're welcome 👍

PR #103, opened 2016-12-02
URL: use escaped argument separator for query

Slamdunk:
During the `assemble()` of a url in a TreeRouteStack router, the query part is passed to a Uri object:
https://github.com/zendframework/zend-router/blob/release-3.0.2/src/Http/TreeRouteStack.php#L393-L395
https://github.com/zendframework/zend-uri/blob/release-2.5.2/src/Uri.php#L788-L798

The Uri object uses the built-in php function `http_build_query`, which leans on the default `&` parameter to separate query arguments.

This is perfectly fine in an general abstract context, but is not correct in a view context where url should be encoded, and so should be its separator, resulting in a `&amp;`.

### DIFF
--- a/src/Helper/Url.php
+++ b/src/Helper/Url.php
@@ -103,6 +103,10 @@ class Url extends AbstractHelper
 
         $options['name'] = $name;
 
+        if (isset($options['query'])) {
+            $options['query'] = str_replace('+', '%20', http_build_query($options['query'], '', '&amp;'));
+        }
+
         return $this->router->assemble($params, $options);
     }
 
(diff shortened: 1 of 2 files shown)


Ocramius:
This is a BC break, since most applications pass the results of the URL
helper to the escaper.
The URL helper also isn't HTML specific.

Slamdunk:
I always used escapers, but due to the fact that in the documentation the UrlHelper is used alone:
https://zendframework.github.io/zend-view/helpers/url/
I always thought it was automatically escaped.

Can you confirm me that it is supposed to be unescaped, and so I can PR a doc change with a note that UrlHelper output should be manually escaped if used as it is?

Ocramius:
@Slamdunk it is unescaped, so we probably need to fix the docs...

PR #106, opened 2017-02-09
Asset helper for various resourses.

turrsis:
The `Assets` helper allow to add various resourses to html page, such as .css, .js, etc.
It can be configure in config files.
For page performance optimisation assets can be placed in top or bottom of page (or elsewhere).

Allow to build several link by one alias:
```
// CONFIG FILE:
'assets_manager' => [
  ...
  'several_assets' => [
    'assets' => [
      'foo.css',
      'bar.css',
      'jquery.js' => 'https://code.jquery.com/jquery-3.0.0.min.js',
    ],
  ],
  ...

// LAYOUT:
echo $this->assets()->add('several_assets');

// OUTPUT:
<link href="/foo.css" type="text/css">
<link href="/bar.css" type="text/css">
<script type="application/javascript" src="https://code.jquery.com/jquery-3.0.0.min.js"></script>
```

Also can possible change layout decoration 'on fly' :
```
// CONFIG FILE:
'assets_manager' => [
  'assets' => [
    'default' => [
      'layout.css' => 'foo.css',
    ],
    'my-theme' => [
      'layout.css' => 'bar.css',
    ],
  ...

// LAYOUT:
$this->assets()->add('layout.css');
if ('good weather') {
  $this->assets()->getAssetsManager()->setCurrentGroup('my-theme');
} else {
  $this->assets()->getAssetsManager()->setCurrentGroup('default');
}
// GOOD WEATHER OUTPUT:
<link href="/bar.css" type="text/css">
// NOT GOOD WEATHER OUTPUT:
<link href="/foo.css" type="text/css">

```

### DIFF
--- a/doc/book/helpers/assets.md
+++ b/doc/book/helpers/assets.md
@@ -1,0 +1,160 @@
+# Assets
+
+The `Assets` helper allow to add various resourses to html page, such as .css, .js, etc.
+It can be configure in config files.
+For page performance optimisation assets can be placed in top or bottom of page (or elsewhere).
+Allow to build several link by one alias.
+Also can possible change layout decoration 'on fly'.
+
+## Basic Usage
+
+```php
+<?php
+// setting links in a view script:
+$this->assets()
+        ->add('http://com.com/foo.css')
+        ->add('/bar.css')
+        ->add('/bat.js');
+
+// rendering the links from the layout:
+echo $this->assets();
+?>
+```
+
+Output:
+
+```html
+<link href="http://com.com/foo.css" type="text/css">
+<link href="/bar.css" type="text/css">
+<script type="application/javascript" src="/bat.js"></script>
+```
+
+## Add assets to different positions on page
+
+```php
+<?php
+// setting links in a view script:
+$this->assets()
+        ->addHeader('/foo.css')
+        ->addFooter('/bar.css');
+
+// rendering the links from the layout:
+echo '<body>';
+    echo '<header>';
+    echo $this->assets()->renderHeader();
+    echo '</header>';
+    echo '<article> some content </article>';
+    echo '<footer>';
+    echo $this->assets()->renderFooter();
+    echo '</footer>';
+echo '</body>';
+?>
+```
+
+Output:
+
+```html
+<body>
+    <header>
+    <link href="/foo.css" type="text/css">
+    </header>
+    <article> some content </article>
+    <footer>
+    <link href="/bar.css" type="text/css">
+    </footer>
+</body>
+```
+
+## Using configurable assets
+
+```php
+<?php
+// section in config file:
+'assets_manager' => [
+    'assets' => [
+        'default' => [
+            'external.css' => [
+                'source' => 'http://com.com/foo.css',
+            ],
+            'foo.css' => [],
+            'several_assets' => [
+                'assets' => [
+                    'external.css',
+                    'bar.css',
+                    'bat.js' => [
+                        'attributes' => [
+                            'charset' => 'UTF-8',
+                        ],
+                    ],
+                ],
+            ],
+        ],
+    ],
+];
+?>
+```
+
+```php
+<?php
+// setting links in a view script:
+$this->assets()
+        ->add('foo.css')
+        ->add('several_assets');
+
+// rendering the links from the layout:
+echo $this->assets();
+?>
+```
+
+Output:
+
+```html
+<link href="/foo.css" type="text/css">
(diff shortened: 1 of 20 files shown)


froschdesign:
In times of "CSS pre-processor", "PostCSS" and Co. is this really needed?

With your view helper and MVC extension you introduce a new heavyweight:

- own cache solution
- new resolver, manager etc.
- hard dependency to the `zend-mvc`component
- …

---

Please have a look at this light-weight solution: #64

weierophinney:
This looks **great**. However, I agree with @froschdesign that this may be a bit much to add directly to zend-view, due to the number of classes required; it's quite a bit of maintenance.

Might I suggest adding this as a separate repository/package instead? It could expose a `ConfigProvider` and/or `Module` class to ensure that the helper is wired when installed, and then have it's own development cycles separate from zend-view.

Thoughts?

turrsis:
This helper not require MVC extention, only Router. But `Url` helper use router too.
I'm not entirely convinced of the necessity of the https://github.com/zendframework/zend-mvc/pull/216.
However, this PR can help to develop own cache, filter and other functionality.
There is a great analogue : https://github.com/RWOverdijk/AssetManager, but, to my mind, it too heavy for simple tasks. The idea here is approximately the same.
In my view, this PR will make the `zend-view` more flexible.

turrsis:
I would suggest to look to replace (perhaps) the `HeadLink`, `HeadStyle` and `HeadScript` this helper.

froschdesign:
> This helper not require MVC extention, only Router.

Your PR on the `zend-mvc` says something different. 😉 

> There is a great analogue : https://github.com/RWOverdijk/AssetManager

Ah, there is the _separate_ package. 

---

I'm not against the idea, but I would prefer a solution which can be used in a `zend-mvc` and a `zend-expressive` application. (Or in a application which only uses `zend-view`.)

Btw. this reminds me of the [Drupal\Core\Asset](https://github.com/drupal/drupal/tree/8.4.x/core/lib/Drupal/Core/Asset)

turrsis:
`zend-mvc-PR` require `zend-view-PR`, but `zend-view-PR` not require `zend-mvc-PR`. If this wrong - this is my mistake and I will fix it. This PR was planned as standalone of `zend-mvc`.

Unfortunately, but I don't know anything about `zend-expressive`.

turrsis:
Both PR were moved to the module https://github.com/turrsis/zend-view-assets.
Can you review it?

turrsis:
@froschdesign, @weierophinney https://github.com/turrsis/zend-view-assets is refactored.
`Mvc` folder should be moved to `zend-mvc` and can be used for mvc applications.
Other classes depends on `zend-view`, `zend-cache`, `zend-filter` and can be used for application which only uses `zend-view`.
Using in `zend-expressive` require additional PR for `zend-expressive` module.
If this is acceptable to you - I will move `Mvc` folder to https://github.com/zendframework/zend-mvc/pull/216.

froschdesign:
@turrsis 
Have you seen the [new `Asset` helper](https://docs.zendframework.com/zend-view/helpers/asset/)?

And do not forget the @weierophinney comment:

> Might I suggest adding this as a separate repository/package instead? It could expose a ConfigProvider and/or Module class to ensure that the helper is wired when installed, and then have it's own development cycles separate from zend-view.

The separate repository is a very good suggestion. No need to add some classes under different repositories / packages.

turrsis:
@froschdesign, @weierophinney 
Yes, I see https://docs.zendframework.com/zend-view/helpers/asset, but it only map resources and not allow use resources from modules.
`turrsis/zend-view-assets` allow to use resources from modules, group and filter it.
"The separate repository is a very good suggestion" - I done it as separate repository, but `mvc` part should be move from this. Where better to move it?
Can this module be included in the zendframework?

froschdesign:
@turrsis 
> but it only map resources and not allow use resources from modules.

Right, but I see a problem here: two view helpers with the equal / similar name.

> turrsis/zend-view-assets allow to use resources from modules, group and filter it.

The modules depending on zend-mvc. I can not use this with [modules in zend-expressive](https://docs.zendframework.com/zend-expressive/features/modular-applications/). Right?

turrsis:
@froschdesign 
Can be renamed to `Resources`.
`Mvc` dependency was moved to https://github.com/turrsis/zend-view-assets-mvc.
Now `turrsis/zend-view-assets` is not depending on `zend-mvc` and (to my mind) can be used with 'zend-expressive'.

weierophinney:
Closing, as the third party module has been created as requested.

PR #110, opened 2017-03-10
Added fallback for navigation container names in view helpers

froschdesign:
- calling the navigation helper with a parameter `navigation` is confusing
- name do not corresponds to the documentation
- old name `default` can be used

Related to an issue of zend-navigation: zendframework/zend-navigation#43


weierophinney:
Thanks, @froschdesign; pushed to develop for release with 2.9.0

froschdesign:
@weierophinney 
Always a pleasure!

PR #112, opened 2017-03-20
PhpRendererStrategy should be able to handle an event w/o a response

MidnightDesign:
ViewEvent#getResponse() may return null, but PhpRendererStrategy#injectResponse() didn't handle that case.

### DIFF
--- a/src/Strategy/PhpRendererStrategy.php
+++ b/src/Strategy/PhpRendererStrategy.php
@@ -103,12 +103,12 @@ class PhpRendererStrategy extends AbstractListenerAggregate
     public function injectResponse(ViewEvent $e)
     {
         $renderer = $e->getRenderer();
-        if ($renderer !== $this->renderer) {
+        $response = $e->getResponse();
+        if ($renderer !== $this->renderer || $response === null) {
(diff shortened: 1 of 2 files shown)


MidnightDesign:
I wasn't sure what to test for since I just expect the method [not to throw an error](https://github.com/zendframework/zend-view/pull/112/files#diff-384b5bf736d0965c9d28df52c30dd038L125). Please let me know if I should change the test.

MidnightDesign:
What should I do about the PHPUnit incompatibility? (`expectException()`/`setExpectedException()`)

froschdesign:
@MidnightDesign 
> What should I do about the PHPUnit incompatibility?

At the moment `zend-view` still using PHPUnit 4.5.

MidnightDesign:
@froschdesign So I shouldn't do anything and the failing builds are OK?

weierophinney:
@MidnightDesign I've merged other commits today that upgrade the PHPUnit install; as long as your new test passes, it's fine.

weierophinney:
Thanks, @MidnightDesign

PR #114, opened 2017-03-20
Upgrade to PHP 5.6 and PHPUnit 5.7/6.0

weierophinney:
This patch updates the package to:

- Set the minimum supported PHP version to 5.6.
- Set the supported PHPUnit version to either 5.7 or 6.0, depending on PHP version used.

A number of test changes were required:

- All references to `PHPUnit_Framework_TestCase` were updated to the namespaced version. In any case where the class was not imported, an import statement was created.
- All `setExpectedException()` calls were updated to `expectException()`, and, if an expected message was provided, `expectExceptionMessage()`
- A number of `getMock()` calls still existed without usage of `getMockBuilder()`; these were updated.
- A zend-servicemanager test trait for testing container compatibility between zend-servicemanager versions differs from v2 to v3; the two instances in this particular suite were updated to override the necessary tests to vary strategy based on which PHP version was detected.
- New test failures were discovered in the flash messenger tests due to differences in how PHPUnit handles global state; these were corrected.

### DIFF
--- a/test/Helper/AbstractTest.php
+++ b/test/Helper/AbstractTest.php
@@ -9,7 +9,7 @@
 
 namespace ZendTest\View\Helper;
 
-use PHPUnit_Framework_TestCase as TestCase;
+use PHPUnit\Framework\TestCase;
 use ZendTest\View\Helper\TestAsset\ConcreteHelper;
 
 /**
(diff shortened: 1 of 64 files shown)


test/Helper/AbstractTest.php
```diff
@@ -30,7 +30,7 @@ public function setUp()
 
     public function testViewSettersGetters()
     {
-        $viewMock = $this->getMock('Zend\View\Renderer\RendererInterface');
+        $viewMock = $this->getMockBuilder('Zend\View\Renderer\RendererInterface')->getMock();
```
kokspflanze: `createMock` would be easyer, than `getMockBuilder`

test/Helper/CycleTest.php
```diff
@@ -17,7 +17,7 @@
  * @group      Zend_View
  * @group      Zend_View_Helper
  */
-class CycleTest extends \PHPUnit_Framework_TestCase
+class CycleTest extends \PHPUnit\Framework\TestCase
```
kokspflanze: maybe add that directly in the `use` section

PR #115, opened 2017-03-20
phpunit added in use section

kokspflanze:
from my comment in https://github.com/zendframework/zend-view/pull/114


weierophinney:
Thanks, @kokspflanze

Issue #120, opened 2017-04-21
PhpRenderer renders previous content when false is provided


porebskk:
Hello,

First thanks for this wonderful framework, I appreciate the effort everybody is putting into this project. Today i encounterd an unexpected behaviour. Allthough its not part of PhpRenderers::render signature, the renderer will return the previously generated content when false is provided as $nameOrModel parameter.
I would expect that the renderer is trying to be as stateless as possible. Is this something that can be categorized as misuse of the class or a behaviour that needs fixing?  
The same output would occur if a parameter would be provided that counts as falsy see:
`while ($this->__template = array_pop($this->__templates)) {`  
  
If this is something that needs fixing I could provide tests and a fix.

Kind regards

weierophinney:
@porebskk Could you at the least create a pull request with a pull request demonstrating the issue? That would then allow us to verify/validate; if you include a fix, that also allows us to validate the fix against the test.

Thanks!

PR #121, opened 2017-04-26
Test for ensuring, that the PhpRenderer doesnt use previous rendered output

porebskk:
See Issue https://github.com/zendframework/zend-view/issues/120

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ cache:
 
 env:
   global:
+    - COMPOSER_PROCESS_TIMEOUT=600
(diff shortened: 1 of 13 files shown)


weierophinney:
@porebskk I've pushed a commit that fixes the issue, but need to wait for Travis to run the tests before merging.

Thanks for the excellent unit tests!

weierophinney:
Thanks, @porebskk, for the great unit test! I pushed the fix to your branch, and it has now been verified with continuous integration; I will release it soon with version 2.9.1.

PR #123, opened 2017-05-17
Return early on HelperPluginManager::injectTranslator() when helper has translator already

bacinsky:
When creating helper via factory with injected translator the HelperPluginManager overrides this translator with the default one, so check is added to prevent this.


weierophinney:
Thanks, @bacinsky; cherry-picked to master for release with 2.9.1.

PR #125, opened 2017-06-04
PHP 7 - Php Renderer - Catching Error - Bug fix

aft-christophe:
The php renderer was only catching Exception - PHP 7 changed how error are reported 
http://php.net/manual/en/language.errors.php7.php


weierophinney:
Thanks, @aft-christophe!

PR #131, opened 2017-08-09
 replace count() usage to empty() when possible

samsonasik:
ref: http://maettig.com/code/php/php-performance-benchmarks.php#empty

### DIFF
--- a/src/Helper/HeadLink.php
+++ b/src/Helper/HeadLink.php
@@ -387,7 +387,7 @@ class HeadLink extends Placeholder\Container\AbstractStandalone
             return false;
         }
 
-        if (! empty($args)) {
+        if ($args) {
             $media = array_shift($args);
             if (is_array($media)) {
                 $media = implode(',', $media);
@@ -395,7 +395,7 @@ class HeadLink extends Placeholder\Container\AbstractStandalone
                 $media = (string) $media;
             }
         }
-        if (! empty($args)) {
+        if ($args) {
             $conditionalStylesheet = array_shift($args);
             if (! empty($conditionalStylesheet) && is_string($conditionalStylesheet)) {
                 $conditionalStylesheet = (string) $conditionalStylesheet;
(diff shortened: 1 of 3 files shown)


src/Helper/HeadLink.php
```diff
@@ -387,15 +387,15 @@ public function createDataStylesheet(array $args)
             return false;
         }
 
-        if (0 < count($args)) {
+        if (! empty($args)) {
```
Ocramius: All these `! empty($args)` can be simplified to `$args` in the context of boolean expressions

src/Helper/HeadLink.php
```diff
@@ -387,15 +387,15 @@ public function createDataStylesheet(array $args)
...
             $media = array_shift($args);
             if (is_array($media)) {
                 $media = implode(',', $media);
             } else {
                 $media = (string) $media;
             }
         }
-        if (0 < count($args)) {
+        if (! empty($args)) {
```
Ocramius: same

src/Helper/HeadLink.php
```diff
@@ -404,7 +404,7 @@ public function createDataStylesheet(array $args)
             }
         }
 
-        if (0 < count($args) && is_array($args[0])) {
+        if (! empty($args) && is_array($args[0])) {
```
Ocramius: same

src/Helper/HeadLink.php
```diff
@@ -452,7 +452,7 @@ public function createDataAlternate(array $args)
         $type  = array_shift($args);
         $title = array_shift($args);
 
-        if (0 < count($args) && is_array($args[0])) {
+        if (! empty($args) && is_array($args[0])) {
```
Ocramius: same

src/Model/ViewModel.php
```diff
@@ -365,7 +365,7 @@ public function getChildren()
      */
     public function hasChildren()
     {
-        return (0 < count($this->children));
+        return (! empty($this->children));
```
Ocramius: `return (bool) $this->children;`

samsonasik:
@Ocramius done ;)

Ocramius:
LGTM :+1:

PR #133, opened 2017-08-14
Don't show prefix and postfix if there is no placeholder items.

AndyDune:
For show h1 in layout with set in view file.
```
<?= $this->placeholder('h1')->setPrefix('<h1>')->setPostfix('</h1>'); ?>
```

### DIFF
--- a/src/Helper/Placeholder/Container/AbstractContainer.php
+++ b/src/Helper/Placeholder/Container/AbstractContainer.php
@@ -117,14 +117,17 @@ abstract class AbstractContainer extends ArrayObject
             ? $this->getWhitespace($indent)
             : $this->getIndent();
 
-        $items  = $this->getArrayCopy();
         // If we don't have items - do not show prefix and postfix
-        if (!count($items)) {
+        if (!$this->count()) {
             return '';
         }
+
+        $itemsToString = implode($this->getSeparator(), $this->getArrayCopy());
+        // todo check empty $itemsToString after trim() to return ''
+
         $return = $indent
             . $this->getPrefix()
-            . implode($this->getSeparator(), $items)
+            . $itemsToString
             . $this->getPostfix();
         $return = preg_replace("/(\r\n?|\n)/", '$1' . $indent, $return);
 
--- a/src/Helper/Placeholder/Container/AbstractContainer.php
+++ b/src/Helper/Placeholder/Container/AbstractContainer.php
@@ -113,8 +113,7 @@ abstract class AbstractContainer extends ArrayObject
      */
     public function toString($indent = null)
     {
-        // If we don't have items - do not show prefix and postfix
-        if (! $this->count()) {
+        if (! count($this)) {
             return '';
         }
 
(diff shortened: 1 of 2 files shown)


src/Helper/Placeholder/Container/AbstractContainer.php
```diff
@@ -118,6 +118,10 @@ public function toString($indent = null)
             : $this->getIndent();
 
         $items  = $this->getArrayCopy();
+        // If we don't have items - do not show prefix and postfix
```
froschdesign: If no item is present, then this check should the first step in this method. You can use `$this->count()`.

froschdesign: Please provide also an unit test for this new behaviour. Thanks!

src/Helper/Placeholder/Container/AbstractContainer.php
```diff
@@ -117,14 +117,17 @@ public function toString($indent = null)
             ? $this->getWhitespace($indent)
             : $this->getIndent();
 
-        $items  = $this->getArrayCopy();
         // If we don't have items - do not show prefix and postfix
-        if (!count($items)) {
+        if (!$this->count()) {
```
froschdesign: Please move this check before `$indent = ($indent !== null)` (line 116) and add a whitespace after `!`.

src/Helper/Placeholder/Container/AbstractContainer.php
```diff
@@ -117,14 +117,17 @@ public function toString($indent = null)
...
             return '';
         }
+
+        $itemsToString = implode($this->getSeparator(), $this->getArrayCopy());
+        // todo check empty $itemsToString after trim() to return ''
```
froschdesign: If items with whitespace are added, then there is a reason for it and we should allow the rendering. Please undo this changes.

src/Helper/Placeholder/Container/AbstractContainer.php
```diff
@@ -113,21 +113,19 @@ public function __toString()
...
-            : $this->getIndent();
-
         // If we don't have items - do not show prefix and postfix
         if (!$this->count()) {
             return '';
         }
 
-        $itemsToString = implode($this->getSeparator(), $this->getArrayCopy());
-        // todo check empty $itemsToString after trim() to return ''
+        $indent = ($indent !== null)
```
malukenho: May be we should avoid the negative check:

```php
$indent = $indent === null
    ? $this->getIndent()
    : $this->getWhitespace($indent);
```

src/Helper/Placeholder/Container/AbstractContainer.php
```diff
@@ -113,8 +113,7 @@ public function __toString()
      */
     public function toString($indent = null)
     {
-        // If we don't have items - do not show prefix and postfix
```
froschdesign: The comment was good.

weierophinney:
Thanks, @AndyDune; merged to develop for release with 2.10.0, as it is a slight change in behavior.

PR #137, opened 2017-08-30
Fixes #134 : add Exception on HeadeMeta::setCharset() when the doctype is Xhtml

### LINKED ISSUE #134: Setting charset in Doctype::XHTML1_RDFA fails
I'm getting an error when I set the charset on a rdfa document. Is that expected? Or an oversight?

```
        // set doctype
        $this->helperPluginManager->get('doctype')
            ->setDoctype(\Zend\View\Helper\Doctype::XHTML1_RDFA);

$this->helperPluginManager->get('headMeta')
            ->setCharset('utf-8');
```

```
Invalid value passed to set; please use setMeta()
./vendor/zendframework/zend-view/src/Helper/HeadMeta.php:442
```

Further; following the suggestion in the error gives me an error too.

```
$this->helperPluginManager->get('headMeta')
         ->setMeta('charset', 'utf-8');
```

```
Method "setMeta" does not exist
./vendor/zendframework/zend-view/src/Helper/Placeholder/Container/AbstractStandalone.php:83
```

samsonasik:
Fixes #134

### DIFF
--- a/src/Helper/HeadMeta.php
+++ b/src/Helper/HeadMeta.php
@@ -367,7 +367,7 @@ class HeadMeta extends Placeholder\Container\AbstractStandalone
     /**
      * Append
      *
-     * @param  string $value
+     * @param  object $value
      * @return void
      * @throws Exception\InvalidArgumentException
      */
@@ -375,7 +375,7 @@ class HeadMeta extends Placeholder\Container\AbstractStandalone
     {
         if (! $this->isValid($value)) {
             throw new Exception\InvalidArgumentException(
-                'Invalid value passed to append; please use appendMeta()'
+                'Invalid value passed to append'
             );
         }
 
@@ -420,7 +420,7 @@ class HeadMeta extends Placeholder\Container\AbstractStandalone
     /**
      * Prepend
      *
-     * @param  string $value
+     * @param  object $value
      * @throws Exception\InvalidArgumentException
      * @return void
      */
@@ -428,7 +428,7 @@ class HeadMeta extends Placeholder\Container\AbstractStandalone
     {
         if (! $this->isValid($value)) {
             throw new Exception\InvalidArgumentException(
-                'Invalid value passed to prepend; please use prependMeta()'
+                'Invalid value passed to prepend'
             );
         }
 
@@ -438,14 +438,14 @@ class HeadMeta extends Placeholder\Container\AbstractStandalone
     /**
      * Set
      *
-     * @param  string $value
+     * @param  object $value
      * @throws Exception\InvalidArgumentException
      * @return void
      */
     public function set($value)
     {
         if (! $this->isValid($value)) {
-            throw new Exception\InvalidArgumentException('Invalid value passed to set; please use setMeta()');
+            throw new Exception\InvalidArgumentException('Invalid value passed to set');
         }
 
         $container = $this->getContainer();
--- a/src/Helper/HeadMeta.php
+++ b/src/Helper/HeadMeta.php
@@ -458,10 +458,15 @@ class HeadMeta extends Placeholder\Container\AbstractStandalone
      * Not valid in a non-HTML5 doctype
      *
      * @param  string $charset
+     * @throws Exception\RuntimeException
      * @return HeadMeta Provides a fluent interface
      */
     public function setCharset($charset)
     {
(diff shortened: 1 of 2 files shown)


src/Helper/HeadMeta.php
```diff
@@ -458,10 +458,15 @@ public function set($value)
      * Not valid in a non-HTML5 doctype
      *
      * @param  string $charset
+     * @throws Exception\RuntimeException
      * @return HeadMeta Provides a fluent interface
      */
     public function setCharset($charset)
     {
+        if ($this->view->plugin('doctype')->isXhtml()) {
+            throw new Exception\RuntimeException('XHTML* doctype has no attribute charset');
```
froschdesign: Please move this check to the `isValid` method and call `isValid` before `$this->set($item);` in the current method.

froschdesign: You should also extend the exception message:

> 'XHTML* doctype has no attribute charset; please use appendHttpEquiv'

Then change the exception class to `Exception\InvalidArgumentException`.

froschdesign:
@samsonasik 
The bug report includes another problem:

> `->setMeta('charset', 'utf-8');`
> > Method "setMeta" does not exist

The exception message is wrong:

https://github.com/zendframework/zend-view/blob/4d9954c05a3981ca711bd55932dc1d2b55580392/src/Helper/HeadMeta.php#L442

Remove the hint to the method "setMeta" and update the DocBlock. Add a short example and fix also the data type for the value parameter.
The same for the methods `prepend` and `append`.

Thanks in advance!

src/Helper/HeadMeta.php
```diff
@@ -334,22 +335,27 @@ protected function isValid($item)
             return false;
         }
 
+        $doctype = $this->view->plugin('doctype');
+        if ($item->type === 'charset' && $doctype->isXhtml()) {
+            throw new Exception\RuntimeException('XHTML* doctype has no attribute charset; please use appendHttpEquiv()');
```
froschdesign: `return false` here.

src/Helper/HeadMeta.php
samsonasik: updated

src/Helper/HeadMeta.php
```diff
@@ -458,20 +464,17 @@ public function set($value)
...
-        if ($this->view->plugin('doctype')->isXhtml()) {
-            throw new Exception\RuntimeException('XHTML* doctype has no attribute charset');
-        }
-
         $item = new stdClass;
         $item->type = 'charset';
         $item->charset = $charset;
         $item->content = null;
         $item->modifiers = [];
+        $this->isValid($item);
```
froschdesign: If false, then throw `Exception\InvalidArgumentException` here.

froschdesign: Compare with the methods `set`, `prepend` etc.

samsonasik:
I've updated exception messages and docblock param in HeadMeta::set(), append(), prepend(), do you have suggestion for data type parameter ? the stdClass check already in `isValid()` method so I think the type parameter should not be used, instead only update the docblock parameter.

src/Helper/HeadMeta.php
samsonasik: already throw an Exception, need a return false ?

samsonasik: understood, make a RuntimeException in `setCharset()` ?

src/Helper/HeadMeta.php
samsonasik: updated

froschdesign:
> …instead only update the docblock parameter.

That's what i meant!

src/Helper/HeadMeta.php
```diff
@@ -474,7 +473,11 @@ public function setCharset($charset)
         $item->charset = $charset;
         $item->content = null;
         $item->modifiers = [];
-        $this->isValid($item);
+
+        if (! $this->isValid($item)) {
+            throw new Exception\RuntimeException('XHTML* doctype has no attribute charset; please use appendHttpEquiv()');
```
froschdesign: You don't like the `InvalidArgumentException`?! 😉

samsonasik: updated to InvalidArgumentException

froschdesign:
@samsonasik 
Thanks for your contribution! 👍

PR #138, opened 2017-09-14
Allow to use ‘as’ attribute on HTML ‘link’ element

gscscnd:
This patch allows to use the `as` attribute by the [`HeadLink` helper](https://github.com/zendframework/zend-view/blob/master/src/Helper/HeadLink.php).

Citing from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content):

> Using `as` to specify the type of content to be preloaded allows the browser to:
>
> * Prioritize resource loading more accurately.
> * Match future requests, reusing the same resource if appropriate.
> * Apply the correct [content security policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) to the resource.
> * Set the correct [`Accept`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) request headers for it.

### DIFF
--- a/src/Helper/HeadLink.php
+++ b/src/Helper/HeadLink.php
@@ -51,7 +51,8 @@ class HeadLink extends Placeholder\Container\AbstractStandalone
         'extras',
         'itemprop',
         'crossorigin',
-        'integrity'
+        'integrity',
+        'as'
     ];
 
     /**
--- a/src/Helper/HeadLink.php
+++ b/src/Helper/HeadLink.php
@@ -52,7 +52,7 @@ class HeadLink extends Placeholder\Container\AbstractStandalone
         'itemprop',
         'crossorigin',
         'integrity',
-        'as'
+        'as',
     ];
 
     /**
--- a/test/Helper/HeadLinkTest.php
+++ b/test/Helper/HeadLinkTest.php
@@ -457,4 +457,10 @@ class HeadLinkTest extends TestCase
         $this->helper->prependAlternate(['itemprop' => 'url', 'href' => '/bar/baz', 'rel' => 'canonical']);
         $this->assertContains('itemprop="url"', $this->helper->toString());
     }
+
+    public function testAsAttributeIsSupported()
+    {
+        $this->helper->headLink(['as' => 'style', 'href' => '/foo/bar.css', 'rel' => 'preload']);
+        $this->assertContains('as="style"', $this->helper->toString());
+    }
 }


gscscnd:
@froschdesign, do you think that a unit test similar to https://github.com/zendframework/zend-view/blob/master/test/Helper/HeadLinkTest.php#L437-L444 would be enough?

froschdesign:
@gscscnd 
> This patch allows to use the as attribute by the HeadLink helper.

The preload links are still work in progress. ([Working Draft 2017-08-30](https://www.w3.org/TR/preload/))
We should not add this new HTML attribute, because it's too early and changes are possible.

We leave the issue open and add the label "WIP".

gscscnd:
Ok, thank you for the explanation.

froschdesign:
> do you think that a unit test similar to https://github.com/zendframework/zend-view/blob/master/test/Helper/HeadLinkTest.php#L437-L444 would be enough?

Right!

src/Helper/HeadLink.php
```diff
@@ -51,7 +51,8 @@ class HeadLink extends Placeholder\Container\AbstractStandalone
         'extras',
         'itemprop',
         'crossorigin',
-        'integrity'
+        'integrity',
+        'as'
```
froschdesign: Please add a comma after `'as'`.

thomasvargiu:
@froschdesign I need it too!

[rfc5988](https://tools.ietf.org/html/rfc5988) is not in draft stage anymore (candidate).

If browsers started implementing this feature, we should provide it too IMO.

froschdesign:
@gscscnd 
> I need it too!

You can always extend the view helper or write your own. This allows you to use the attribute _now_.

> If browsers started implementing this feature…

Do you have a status overview which includes this new attribute? Thanks!

thomasvargiu:
You’re right, I can always use my own view helper.

About browser compatibility you can see it [here](https://caniuse.com/#feat=link-rel-preload)

froschdesign:
Next step is reached: "[HTML 5.2 - W3C Recommendation, 14 December 2017](https://www.w3.org/TR/html52/)"

froschdesign:
Final step is reached: "[HTML 5.2 is done](https://www.w3.org/blog/2017/12/html-5-2-is-done-html-5-3-is-coming/)"

This will be added with the next minor release. When? I don't know. At the moment I work on some other task. I will try to clarify this.

Thanks for your patience!

MatthiasKuehneEllerhold:
Any news on this? "as" is a critical part of the HTTP/2 Server Push via Apache http2 module.

weierophinney:
Thanks, @gscscnd; merged to develop for release with 2.10.0.

PR #143, opened 2017-12-16
type attribute for the style element is not needed for HTML5

matbech:
The type attribute is not required for style.

This fixes the following warning from the HTML validator:
 The type attribute for the style element is not needed and should be omitted.


froschdesign:
@matbech 
> The type attribute is not required for style.

This is only true for [HTML5](https://www.w3.org/TR/html5/document-metadata.html#the-style-element), not for [XHTML](https://www.w3.org/TR/xhtml1/dtds.html#dtdentry_xhtml1-strict.dtd_style) and [HTML4](https://www.w3.org/TR/html401/present/styles.html#h-14.2.3).

froschdesign:
Thanks for the changes.
Now we need unit tests for this new behaviour. Check with and without the doctype HTML5.

weierophinney:
Closing, as the requested changes made ~1year ago have not been made.

PR #146, opened 2018-01-17
sort-packages configuration expects a boolean in latest composer versions

weierophinney:
With the 1.6 series of composer, `sort-packages` MUST be a boolean value, or the `composer.json` will not validate, blocking builds.


weierophinney:
Fixed with #121.

PR #153, opened 2018-04-19
Fluent Interface missing on ConsoleModel::setErrorLevel

porebskk:
Hello,

contrary to almost all other setter methods on classes that implement the ModelInterface, ConsoleModel::setErrorLevel ist not returning itself when the method is called.
I would assume that it would return $this.

Best regards

### DIFF
--- a/src/Model/ConsoleModel.php
+++ b/src/Model/ConsoleModel.php
@@ -31,10 +31,12 @@ class ConsoleModel extends ViewModel
      * Set error level to return after the application ends.
      *
      * @param int $errorLevel
+     * @return $this
      */
     public function setErrorLevel($errorLevel)
     {
         $this->options['errorLevel'] = $errorLevel;
+        return $this;
     }
 
     /**
--- a/test/Model/ConsoleModelTest.php
+++ b/test/Model/ConsoleModelTest.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace ZendTest\View\Model;
+
+use PHPUnit\Framework\TestCase;
+use Zend\View\Model\ConsoleModel;
+
+class ConsoleModelTest extends TestCase
+{
+    public function testImplementsModelInterface()
+    {
+        $model = new ConsoleModel();
+        $this->assertInstanceOf('Zend\View\Model\ModelInterface', $model);
+    }
+
+    /**
+     * @group zend-view-152
+     * @see https://github.com/zendframework/zend-view/issues/152
+     */
+    public function testSetErrorLevelIsReturningThis()
+    {
+        $model = new ConsoleModel();
+        $actual = $model->setErrorLevel(0);
+        $this->assertSame($model, $actual);
+    }
+}


test/Model/ConsoleModelTest.php
```diff
@@ -0,0 +1,26 @@
...
+     * @group zend-view-152
+     * @see https://github.com/zendframework/zend-view/issues/152
+     */
+    public function testSetErrorLevelIsReturningThis()
+    {
+        $model = new ConsoleModel();
+        $actual = $model->setErrorLevel(0);
+        $this->assertSame($model, $actual);
+    }
+}
```
thexpand: A new line is missing here, causing the build to fail.

test/Model/ConsoleModelTest.php
```diff
@@ -0,0 +1,26 @@
+<?php
```
thexpand: Add license and copyright.
See here: https://github.com/zendframework/maintainers/blob/abab4a99f3e6d1fac4d7400232ce30ee0bc95249/template/src/ConfigProvider.php#L2-L6

test/Model/ConsoleModelTest.php
```diff
@@ -0,0 +1,26 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Zend\View\Model\ConsoleModel;
+
+class ConsoleModelTest extends TestCase
+{
+    public function testImplementsModelInterface()
+    {
+        $model = new ConsoleModel();
+        $this->assertInstanceOf('Zend\View\Model\ModelInterface', $model);
```
thexpand: It's better to use the class constant, instead of writing the FQN as a string.

src/Model/ConsoleModel.php
```diff
@@ -31,10 +31,12 @@ class ConsoleModel extends ViewModel
      * Set error level to return after the application ends.
      *
      * @param int $errorLevel
+     * @return $this
```
froschdesign: Please use `self` here.

webimpress: @froschdesign why `self`? I think `$this` is the most right here, as we return the same instance, not the new one. I think this is also what is suggested in PSR-5 draft: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md - check the bottom of the document: points 13-15.

froschdesign: @webimpress 
I would like to remove it completely! (`>=7.1`)

webimpress: @froschdesign Removing them is a different story and this is not answer for my question, why you think `self` is more appropriate here ;-) 
As I understand fluent interfaces makes sens in some cases, and not always they are "evil".
https://ocramius.github.io/blog/fluent-interfaces-are-evil/

froschdesign: @webimpress 
No, I mean remove the return type in the DocBlock.

webimpress: @froschdesign I think what was decided in ZF is to remove them (PHPDocs) only if they are not providing any additional information. And in that case, even if we add RTH `: self` still `@return $this` gives us more information about returned object (as described - in quoted before - PSR-5 draft).

froschdesign: Then I will stop the discussion here. It makes no sense without any coding standard.

froschdesign: ping https://github.com/zendframework/zend-coding-standard/pull/5

weierophinney:
Thanks, @porebskk. I made some minor consistency changes to the unit test, and cherry-picked this against develop for release with 2.11.0.

PR #154, opened 2018-05-07
change strstr/substr with strpos when possible

samsonasik:


src/Helper/AbstractHtmlElement.php
```diff
@@ -72,7 +72,7 @@ protected function htmlAttribs($attribs)
         foreach ((array) $attribs as $key => $val) {
             $key = $escaper($key);
 
-            if (('on' == substr($key, 0, 2)) || ('constraints' == $key)) {
+            if ((0 === strpos($key, 'on')) || ('constraints' == $key)) {
```
webimpress: We can remove useless brackets here as well

samsonasik: done ;)

weierophinney:
Thanks, @samsonasik

weierophinney:
Merged to develop for release with 2.11.0

PR #155, opened 2018-06-17
Try to use AuthenticationServiceInterface for the Identity helper

### LINKED ISSUE #33: Improve identity view helper factory
I have a few problems with an identity view helper [factory](https://github.com/zendframework/zend-view/blob/master/src/Helper/Service/IdentityFactory.php).

```
    public function createService(ServiceLocatorInterface $serviceLocator)
    {
        $services = $serviceLocator->getServiceLocator();
        $helper = new Identity();
        if ($services->has('Zend\Authentication\AuthenticationService')) {
            $helper->setAuthenticationService($services->get('Zend\Authentication\AuthenticationService'));
        }
        return $helper;
    }
```
1. First about a retrieving of auth service. Maybe will be better to use [Zend\Authentication\AuthenticationServiceInterface](https://github.com/zendframework/zend-authentication/blob/master/src/AuthenticationServiceInterface.php) instead of **Zend\Authentication\AuthenticationService** as name of the auth service?
2. By default [HelperPluginManager](https://github.com/zendframework/zend-view/blob/master/src/HelperPluginManager.php) doesn't contain main service container, and var **$services** can be null. But it seems in develop branch this factory class already modified.

gscscnd:
See https://github.com/zendframework/zend-mvc-plugin-identity/pull/9

Closes #33

### DIFF
--- a/src/Helper/Service/IdentityFactory.php
+++ b/src/Helper/Service/IdentityFactory.php
@@ -10,6 +10,8 @@
 namespace Zend\View\Helper\Service;
 
 use Interop\Container\ContainerInterface;
+use Zend\Authentication\AuthenticationService;
+use Zend\Authentication\AuthenticationServiceInterface;
 use Zend\ServiceManager\FactoryInterface;
 use Zend\ServiceManager\ServiceLocatorInterface;
 use Zend\View\Helper\Identity;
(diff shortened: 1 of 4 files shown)


src/Helper/Service/IdentityFactory.php
```diff
@@ -33,6 +33,8 @@ public function __invoke(ContainerInterface $container, $name, array $options =
         $helper = new Identity();
         if ($container->has('Zend\Authentication\AuthenticationService')) {
             $helper->setAuthenticationService($container->get('Zend\Authentication\AuthenticationService'));
+        } elseif ($container->has('Zend\Authentication\AuthenticationServiceInterface')) {
```
froschdesign: We can use the FQCN (fully qualified class name) here.

src/Helper/Service/IdentityFactory.php
```diff
@@ -33,6 +33,8 @@ public function __invoke(ContainerInterface $container, $name, array $options =
         $helper = new Identity();
         if ($container->has('Zend\Authentication\AuthenticationService')) {
```
froschdesign: The same here.

froschdesign:
Please use the `develop` branch and add also an unit test for the factory. (Hint: look at your own PR on zend-mvc-plugin-identity 😉)

Thank you in advance!

weierophinney:
Thanks, @gscscnd!

PR #158, opened 2018-07-20
Cloning ViewModel now also clones it's variables if they are an object

### LINKED ISSUE #157: Changing variables in ViewModel clones are propagated
By default, `\Zend\View\Model\ViewModel::$variables` is a `\Zend\View\Variables` instance.
When cloning an instance of `ViewModel`, if `$variables` property is a reference to an object instead of a native php array, calling `setVariables` on the clone will also change the variables in the original `ViewModel` instance.
This becomes a problem when using `zend-view` as a view renderer in Zend Expressive on a long-running process (see https://github.com/zendframework/zend-expressive-zendviewrenderer/blob/2.0.0/src/ZendViewRenderer.php#L319), and causes layout variables to be stuck in the same value between requests.

### How to reproduce:
```php
<?php
use \Zend\View\Model\ViewModel;

$viewModel1 = new ViewModel();
$viewModel1->setVariables(['a' => 'foo']);

$viewModel2 = clone $viewModel1;
$viewModel2->setVariables(['a' => 'bar']);

echo $viewModel1->a;

// outputs 'bar'
// expected output is 'foo'
```

samuelnogueira:
Fixes #157


weierophinney:
Thanks, @samuelnogueira! Cherry-picked to develop for release with 2.11.0.

Issue #16, opened 2015-10-02
Placeholder View Helper "containerExists" makes no sense


montolio:
At the moment the public function containerExists does not have any change to work:

because __invoke checks if a name is set it is impossible to call placeholder()
So to check if a placeholder name exists you have to all call 
placeholder('foobar')->containerExists('foobar')

but then the container is created in __invoke and containerExists returns in every case true.

To avoid this placeholder('dummy')->containerExists('foobar') would work, but that's an ugly workaround!

michalbundyra:
This repository has been closed and moved to laminas/laminas-view; a new issue has been opened at https://github.com/laminas/laminas-view/issues/32.

PR #162, opened 2018-08-15
Fix Placeholder view helper return value

thexpand:
This PR aims to fix the bug, outlined in the following issue: https://github.com/zendframework/zend-view/issues/16

The changes will allow `$this` as a return value of the view helper, when no name is passed as an argument, thus providing access to the methods inside the view helper. For this reason, more tests have been added to `ZendTest\View\Helper\PlaceholderTest`.

### DIFF
--- a/src/Helper/Placeholder.php
+++ b/src/Helper/Placeholder.php
@@ -9,7 +9,6 @@
 
 namespace Zend\View\Helper;
 
-use Zend\View\Exception\InvalidArgumentException;
 use Zend\View\Helper\Placeholder\Container;
 
 /**
@@ -37,15 +36,12 @@ class Placeholder extends AbstractHelper
      * Placeholder helper
      *
      * @param  string $name
-     * @throws InvalidArgumentException
-     * @return Placeholder\Container\AbstractContainer
+     * @return Placeholder\Container\AbstractContainer|Placeholder
      */
     public function __invoke($name = null)
     {
         if ($name === null) {
-            throw new InvalidArgumentException(
-                'Placeholder: missing argument. $name is required by placeholder($name)'
-            );
+            return $this;
         }
 
         $name = (string) $name;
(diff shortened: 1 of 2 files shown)


src/Helper/Placeholder.php
```diff
@@ -37,15 +36,12 @@ class Placeholder extends AbstractHelper
      * Placeholder helper
      *
      * @param  string $name
-     * @throws InvalidArgumentException
-     * @return Placeholder\Container\AbstractContainer
+     * @return Placeholder\Container\AbstractContainer|Placeholder
```
Ocramius: self

src/Helper/Placeholder.php
```diff
@@ -37,15 +36,12 @@ class Placeholder extends AbstractHelper
...
      */
     public function __invoke($name = null)
     {
         if ($name === null) {
-            throw new InvalidArgumentException(
```
Ocramius: This particular change is most likely a BC break. Although I don't think people actually do rely on try-catching it, it is possible that they rely on the page to be crashing

thexpand:
What could we do about the failing Travis build?
If we use `zend-mvc ^2.7` it's failing on PHP 7.2 with `DEPS=lowest`.
If we use `zend-mvc ^2.7.13` it's failing, because of the minimum stability.

src/Helper/Placeholder.php
thexpand: I totally agree with you. Should we move this to the `3.0.0` milestone?

thexpand: @Ocramius ping

froschdesign: @thexpand 
Next major version makes sense, because nobody uses a try-catch-block here.

weierophinney:
This repository has been closed and moved to laminas/laminas-view; a new issue has been opened at https://github.com/laminas/laminas-view/issues/7.

weierophinney:

Issue #107, opened 2017-02-13
meta tag attributes escaped with escapeHtml instead of escapeHtmlAttr


autowp:
headMeta view helper uses [AbstractStandalone::escape](https://github.com/zendframework/zend-view/blob/master/src/Helper/HeadMeta.php#L261) method where [escapeHtml is used](https://github.com/zendframework/zend-view/blob/master/src/Helper/Placeholder/Container/AbstractStandalone.php#L176) instead of escapeHtmlAttr

Also headLink, headScript and other helpers affected.

thexpand:
It's not that easy to just change the `escapeHtml` with `escapeHtmlAttr`, because `AbstractStandalone::escape()` is used not only for attributes. I think a new method `escapeAttribute` could be introduced and will be used accordingly.

Ocramius:
@thexpand I don't understand your comment: why is a new method needed? We know that the meta element always has an attribute with the content.

thexpand:
@Ocramius Because the `escape` method uses the `escapeHtml`, which is okay for HTML content, but not okay for HTML attributes. That's why I think the `escapeHtmlAttr` needs exposure as a new method.

Ocramius:
Ah, I understand - needs two escapers :+1:

PR #164, opened 2018-08-15
Add method for escaping HTML attributes

thexpand:
Covers https://github.com/zendframework/zend-view/issues/107


src/Helper/Placeholder/Container/AbstractStandalone.php
```diff
@@ -176,6 +176,24 @@ protected function escape($string)
         return $this->getEscaper()->escapeHtml((string) $string);
     }
 
+    /**
+     * Escape an attribute value
+     *
+     * @param  string $string
+     * @return string
+     */
+    protected function escapeAttribute($string)
```
Ocramius: Add the return type and parameter type, please. They can be dropped from the docblock

src/Helper/Placeholder/Container/AbstractStandalone.php
```diff
@@ -176,6 +176,24 @@ protected function escape($string)
...
+    {
+        if ($this->getView() instanceof RendererInterface
+            && method_exists($this->getView(), 'getEncoding')
```
Ocramius: What's going on here, exactly?

thexpand: @froschdesign Exactly. I think we could safely drop the if statement and use the return statement only.

src/Helper/Placeholder/Container/AbstractStandalone.php
thexpand: But don't we still support PHP 5.6?
In the `composer.json` there is `"php": "^5.6 || ^7.0"`

Ocramius: URRRRGGGHHHHHHHHHHH @weierophinney any note on that?

weierophinney: Until we bump to a new major version, we support PHP 5.6.

thexpand: Do we leave it like that or...?

Ocramius: What? Since when is a new major needed for version bumps? That's horrible :-\

weierophinney: @Ocramius I've told you this more than once! It was on our blog post announcing the move to 7.1 (https://framework.zend.com/blog/2017-06-06-zf-php-7-1.html):

> - We will start releasing new major versions of components that pin to PHP 7.1+, so we can take advantage of the latest language features.
> - We will no longer test against HHVM. Most of our components were failing against HHVM already, so this does not pose a BC break at this time.
> - We will continue providing security patches only to the last release branch supporting PHP 5.6 until PHP 5.6 reaches its end-of-life date. This allows us to minimize the amount of maintenance we need for supporting these versions, while still providing stable, secure releases for our PHP 5 users.
> - New components will pin to PHP 7.1+.

Only new components, or new major versions of components, are supposed to bump to 7.1. The idea is to allow us to refactor to provide all necessary typehints when we do, which will break BC. Adding typehints is a BC break, thus new major version.

Ocramius: So anything that was in `zendframework/zendframework` is not free to bump, while anything outside of it can bump at each minor?

Ocramius: Note: thinking about `zendframework/zend-code` (for example), which can only really support the latest PHP minor each time

thexpand: @weierophinney If I understand correctly, bumping the minimum PHP version from 5.6 to 7.x is a BC (the major version of the repository must also be changed). However, if we bump the minimum PHP version from 7.1 to 7.2, it's a minor change, so we would only increment the minor version of the repository. Right?

weierophinney: Correct. We have dropped PHP 5.5 support in minor versions throughout the 2.Y series of releases on all components.

Ocramius: Reminder that PHP-SRC does not follow SemVer

weierophinney: Water under the bridge at this point. We adopted this policy because it provides a consistent, understandable message: we will work with new major versions of PHP, but will not adopt their features until we issue a new major release.

When folks see a new 2.Y release come out, but Composer says it cannot update to it, and all bugfixes and security fixes are in that release and forwards, we receive a lot of pushback, as users cannot then update to receive these fixes. Many have then requested having multiple minor release branches for purposes of releasing security fixes — which becomes an even greater maintenance nightmare than we currently have. With a new major release, there is plenty of messaging of changes, users have to opt-in to the new version, and users know they need to update by the end of the support period if they want security fixes.

(For the record, we receive pushback even for bumping a minor PHP version within minor versions! However, it's generally easier to update to a new minor PHP release than to a new major one.)

thexpand: Is there anything else that we can do for this PR?

Ocramius: @thexpand I think PR is fine, sorry for derailing the convo

thexpand: @Ocramius Actually, I'm glad that such a discussion happened :)

thexpand: Side note - is there anything that we can do for the failing Travis build with lowest deps on PHP 7.2? It seems to be happening to almost all PRs of `zend-view`.

thexpand:
@Ocramius I think I've done something wrong.. I tried to rebase onto master, but I believe this is wrong. Maybe I just had to revert the commits from @weierophinney?

Ocramius:
You can cherry-pick your own commits to a new branch and force-push it as
this branch.
Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

thexpand:
Thanks, it worked :+1:
All good now!

DennisDobslaf:
If your ie is set to compatibility mode and you want to send
`<meta http-equiv="X-UA-Compatible" content="IE=edge"/> `
via Zend\View\Helper\HeadMeta like 
`$headMeta->appendHttpEquiv('X-UA-Compatible', 'IE=edge');`
this will lead to the (correct) escaped string
`<meta http-equiv="X-UA-Compatible" content="IE&#x3D;edge">`
But IE will now fall back to IE 7 Mode (in my case). 

Can anyone confirm?

As a workaround I send 
`header('Content-Type: text/html; charset=UTF-8');`
before any other, but that does not feel like the right solution.

froschdesign:
@DennisDobslaf 
There is a open issue for that. Please add your comment there, because this pull request is closed. Thanks!

PR #169, opened 2018-11-12
Fixes #163 : apply zend-json to require

### LINKED ISSUE #163: zend-json required only as dev dependency
Hey there,

why is zend-json only required as dev dependency, although the JsonModel in this package has a dependency on it?

Could it be moved to normal dependencies?

samsonasik:
Fixes #163


weierophinney:
Thanks, @samsonasik

PR #170, opened 2018-11-12
set default extras as empty array to prevent compact() errors

kilip:
Hello,

PHP ```compact()``` function at line 413 will create php error when ```$args``` is an empty array:
https://github.com/zendframework/zend-view/blob/34babb457c8e38c64415b8615de39f9b3b5dcfbc/src/Helper/HeadLink.php#L379-L416

I have found this error in this travis build:
https://travis-ci.org/cross-solution/YAWIK/jobs/453775095#L1805
To reproducing error just simply execute this function ```HeadLink::prependStylesheet('/some/foo.css')```

I think this ```compact()``` error only occurs in php 7.3, but the code is buggy too. We need to define default value for ```$extras``` variables.


src/Helper/HeadLink.php
```diff
@@ -408,6 +408,8 @@ public function createDataStylesheet(array $args)
         if ($args && is_array($args[0])) {
             $extras = array_shift($args);
             $extras = (array) $extras;
+        } else {
```
froschdesign: `else` is not needed here. Move the variable definition before `if` and use `$extras = []`.

src/Helper/HeadLink.php
```diff
@@ -460,6 +462,8 @@ public function createDataAlternate(array $args)
             if (isset($extras['media']) && is_array($extras['media'])) {
                 $extras['media'] = implode(',', $extras['media']);
             }
+        } else {
```
froschdesign: Same here.

froschdesign:
Reference for this change in PHP 7.3: http://php.net/manual/en/function.compact.php#refsect1-function.compact-changelog

froschdesign:
Duplicate of #167

src/Helper/HeadLink.php
```diff
@@ -405,6 +405,7 @@ public function createDataStylesheet(array $args)
             }
         }
 
+        $extras = [];
```
froschdesign: I'm sorry, but can you move this to line 386. Thanks!

src/Helper/HeadLink.php
```diff
@@ -453,6 +454,7 @@ public function createDataAlternate(array $args)
         $type  = array_shift($args);
         $title = array_shift($args);
 
+        $extras = [];
```
froschdesign: And this line under the definition of `$title`.

src/Helper/HeadLink.php
kilip: I have move the other $extra variable too to the default groups. please check

PR #171, opened 2018-11-27
Allow PHP Renderer to Extends other templates.

aft-christophe:
Added a new function at the renderer level so that current template extends others.

e.g

template.phtml
```
<?php 
$this->extendsTemplate('parent_template');
?>

[template content.]
```

parent_template.phtml
```
Above template.phtml
<?php echo $this->content ?>
Under template.phtml
```

Will generate
```
Above template.phtml
[template content.]
Under template.phtml
```


froschdesign:
@aft-christophe 
Have you seen the concept of children for view models?

https://docs.zendframework.com/zend-view/quick-start/#nesting-view-models

aft-christophe:
Hi @froschdesign  
Yes I am aware of nesting view models, but unfortunately it does not work  very well with applications like zend-expressive - Where you you pass a template name to the renderer
https://docs.zendframework.com/zend-expressive/v3/features/template/intro/

I know you could pass the a ViewModel with a child model  as layout but it make more sense to "extends" from the template file itself.
I want to keep my handlers / controllers as detached as possible from my templating renderer (So it can be switched without much headaches).

Similar to how most of Templating systems do it out there
Twig -> https://twig.symfony.com/doc/2.x/tags/extends.html
Blade -> https://laravel.com/docs/5.7/blade#extending-a-layout
Not to sure about plates -> http://platesphp.com/v3/templates/inheritance/

How would you tackle this scenario: 
I want to use a layout AND a parent template :
- template
- parent 
- Layout

Thank you

froschdesign:
@aft-christophe 
Please add these kind of informations always to the description of the pull request otherwise nobody knows the reason or the motivation for your contribution.

aft-christophe:
Closing this PR - 
I was able to achieve same effect by using the FilterChain from the render function instead and Custom View Helpers.

PR #174, opened 2018-12-06
Fixes #93 - Issues with Cycle View Helper docs

### LINKED ISSUE #93: Issues with Cycle View Helper docs
Can someone explain to me what is this: {{{PHP2}}} ?

Also later in the same doc, there is section about:

> You can also cycle in reverse, using the `prev()` method instead of `next()`:

However in the example there is neither a `prev()` nor `next()` method used.

```
<table>
    <?php foreach ($this->books as $book): ?>
    <tr style="background-color: {{{PHP2}}}">
       <td><?php echo $this->escapeHtml($book['author']) ?></td>
    </tr>
    <?php endforeach ?>
</table>
```

How Cycle plugin works then?

froschdesign:


froschdesign:
Looks good: https://docs.zendframework.com/zend-view/helpers/cycle/

PR #176, opened 2018-12-06
Updated Travis CI configuration to pass all builds (replace #173)

webimpress:
Replace #173


weierophinney:
Closing in favor of #173. While I note that the `locked` environments passed... they were not really locked, as they had to add additional dev dependencies. My take is that we simply need to test that it works against zend-mvc v2 and v3; if developers are grabbing the latest zend-view release, they will be on the latest zend-mvc as well in most cases, so a lowest/latest strategy works for us until we can bump to PHP 7.1+ and remove compatibility with zend-mvc v2 releases.

Issue #178, opened 2018-12-07
Require "zend-json": "^2.6.1"


vaclavvanik:
As of version 2.11 zend-json is required. Why is not the constraint like `"zendframework/zend-json": "^2.6.1 || ^3.0"`?

froschdesign:
@vaclavvanik 
Please create a pull request to update the Composer file.

PR #179, opened 2018-12-07
zend-json: "^2.6.1 || ^3.0"

vaclavvanik:
As I noted in #178 this patch allows zend-json: "^2.6.1 || ^3.0".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
     "require": {
         "php": "^5.6 || ^7.0",
         "zendframework/zend-eventmanager": "^2.6.2 || ^3.0",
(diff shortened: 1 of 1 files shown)


webimpress:
Hm... I am not sure what is the point of that PR, as json is the part of PHP core:
http://php.net/manual/en/json.requirements.php

webimpress:
Hm... I mean we should just use ext-json, heh, this is what I wanted to say :)

froschdesign:
@webimpress 
See https://github.com/zendframework/zend-view/issues/163

weierophinney:
Thanks, @vaclavvanik!

Issue #180, opened 2018-12-11
HTML-ENTITIES encoding replace UTF-8 in meta tags


petranek80:
After updating zendframework/zend-view (2.10.0 => 2.11.1) meta tags have HTML-ENTITIES encoding instead of UTF-8

e.g.
<link href="&#x2F;img&#x2F;favicon&#x2F;favicon.ico" rel="shortcut&#x20;icon" type="image&#x2F;vnd.microsoft.icon">

after downgrade to 2.10.0 is back normal, UTF-8 encodings

<link href="/img/favicon/favicon.ico" rel="shortcut icon" type="image/vnd.microsoft.icon">

Ocramius:
@petranek80 can you write a test case that demonstrates the regression, please?

petranek80:
e.g.
`<link href="&#x2F;img&#x2F;favicon&#x2F;favicon.ico" rel="shortcut&#x20;icon" type="image&#x2F;vnd.microsoft.icon">`

after downgrade to 2.10.0 is back normal, UTF-8 encodings

`<link href="/img/favicon/favicon.ico" rel="shortcut icon" type="image/vnd.microsoft.icon">`

Ocramius:
@petranek80 that's not a problem, since the HTML parser will pick up those characters anyway?

av3:
I saw the same as I wanted to check my meta description and OpenGraph information. It's not very easy to read when looking at the source code in the browser. The reason for that behavior is https://github.com/zendframework/zend-view/pull/164/files.

But as you can see there is an option for auto escpaping. I'm now using the following for the output (in my phtml file for the layout:

`<?= $this->headLink()->setAutoEscape(false); ?>`

But you can also use the setAutoEscape method earlier. Unfortunately you can use setAutoEscape() for headMeta(), but it has no effect, because `$this->autoEscape` isn't checked in the HeadMeta.php

hummal:
same here! Not sure it is a problem for search engine crawlers to detect the content properly

froschdesign:
@av3 
> It's not very easy to read when looking at the source code in the browser.

Please check the HTML code in the developer tools of your browser.

froschdesign:
@av3 
> Unfortunately you can use `setAutoEscape()` for `headMeta()`, but it has no effect, because `$this->autoEscape` isn't checked in the `HeadMeta.php`

Can you create a new issue report for this problem? Thanks!

Ocramius:
[...]

PR #182, opened 2019-01-11
Optional Autoescaping in HeadMeta ViewHelper

moegeler:
Bugfixing for https://github.com/zendframework/zend-view/issues/180

### DIFF
--- a/src/Helper/HeadMeta.php
+++ b/src/Helper/HeadMeta.php
@@ -243,7 +243,7 @@ class HeadMeta extends Placeholder\Container\AbstractStandalone
             if (! in_array($key, $this->modifierKeys)) {
                 continue;
             }
-            $modifiersString .= $key . '="' . (($this->autoEscape) ? $this->escapeAttribute($value) : $value) . '" ';
+            $modifiersString .= sprintf('%s="%s"', $key, $this->autoEscape ? $this->escapeAttribute($value) : $value);
         }
 
         $modifiersString = rtrim($modifiersString);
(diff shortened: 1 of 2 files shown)


froschdesign:
@malukenho 
Do you really think that the change from `assertEquals` to `assertNotEquals` is correct? Even if the result is the same and true in both tests? 😉

test/Helper/HeadMetaTest.php
```diff
@@ -619,6 +619,6 @@ public function testConditionalNoIEWidthSpace()
     public function testTurnOffAutoEscapeDoesNotEncode()
     {
         $this->helper->setAutoEscape(false)->appendHttpEquiv('foo', 'bar=baz');
-        $this->assertEquals('<meta http-equiv="foo" content="bar=baz" />', $this->helper->toString());
+        $this->assertNotEquals('<meta http-equiv="foo" content="bar=baz">', $this->helper->toString());
```
froschdesign: This change makes no sense. Please check for the _same_ result.

src/Helper/HeadMeta.php
```diff
@@ -243,7 +243,7 @@ public function itemToString(stdClass $item)
             if (! in_array($key, $this->modifierKeys)) {
                 continue;
             }
-            $modifiersString .= $key . '="' . $this->escapeAttribute($value) . '" ';
+            $modifiersString .= $key . '="' . (($this->autoEscape) ? $this->escapeAttribute($value) : $value) . '" ';
```
froschdesign: Please use `sprintf` for better reading here.

src/Helper/HeadMeta.php
```diff
@@ -271,8 +271,8 @@ public function itemToString(stdClass $item)
         $meta = sprintf(
             $tpl,
             $type,
-            $this->escapeAttribute($item->$type),
-            $this->escapeAttribute($item->content),
+            ($this->autoEscape) ? $this->escapeAttribute($item->$type) : $item->$type,
```
froschdesign: The first brackets can be omitted.

moegeler:
Sorry, that was my first pull-request. I hope that it is alright now.

weierophinney:
Thanks, @moegeler!

PR #188, opened 2019-08-19
Fixes #109 - HeadTitle helper documentation is too complicated

### LINKED ISSUE #109: HeadTitle helper documentation is too complicated
For a basic usage the module class is not needed.

https://docs.zendframework.com/zend-view/helpers/head-title/

froschdesign:


doc/book/helpers/head-title.md
```diff
@@ -19,62 +19,15 @@ explicitly pass a different attach order as the second parameter.
...
 
-You may specify a title tag at any time. A typical usage would have you setting
-title segments for each level of depth in your application: site, module,
-controller, action, and potentially resource. This could be achieved in the
-module class.
+Specify a title tag in a view script, e.g. 
+`module/Album/view/album/album/index.phtml`:
 
 ```php
-// module/MyModule/Module.php
```
webimpress: @froschdesign 
This example has been removed completely.

Do you think it is no longer needed?
Personally I prefer to set it in templates, but maybe we should keep it in chapter "Advanced usage" or something similar? Maybe it will be useful for someone?

froschdesign: @webimpress 
> Do you think it is no longer needed?

This is example is not suitable for practical use because who needs the names of module, controller or action in title?
In addition, the example only conveys that the use of the helper is complicated. But that's not true.

webimpress:
Thanks, @froschdesign!

PR #189, opened 2019-08-20
Fixes #99 - Missing documentation for Escape view helpers

### LINKED ISSUE #99: Missing documentation for Escape view helpers
These helpers needs some explanations and examples:
- `Zend\View\Helper\EscapeCss`
- `Zend\View\Helper\EscapeHtml`
- `Zend\View\Helper\EscapeHtmlAttr`
- `Zend\View\Helper\EscapeJs`
- `Zend\View\Helper\EscapeUrl`

There is only a short notice in the [PhpRenderer section](https://docs.zendframework.com/zend-view/php-renderer/#usage).

froschdesign:


doc/book/helpers/escape.md
```diff
@@ -0,0 +1,248 @@
...
+
+If the object does not contains the methods `__toString()` or `toArray()` then
+the object is casted to an `array`:
+
+```php
+$object = new class {
+    public $headline = '<h1>Foo</h1>';
+};
+
+var_dump($this->escapeHtml($object, Zend\View\Helper\EscapeHtml::RECURSE_OBJECT));
```
webimpress: is the 2nd param required here?
I assume the same result we get without it.

doc/book/helpers/escape.md
```diff
@@ -0,0 +1,248 @@
+# Escape
+
+The following helpers can **escape output in view scripts and defend from XSS
+and related vulnerabilities**. To escape different contexts of a HTML document,
+zend-view provides 5 helpers:
```
webimpress: I don't like fixed numbers, it's a very big chance that it will be not update if the number of escapers change.
Maybe better "provides the following helpers:" ?

doc/book/helpers/escape.md
froschdesign: Without the second parameter an exception is thrown:

https://github.com/zendframework/zend-view/blob/b4861879541dcb7988dcb9b0eb536fc048db098b/src/Helper/Escaper/AbstractHelper.php#L72-L74

doc/book/helpers/escape.md
froschdesign: Very good point! 👍

webimpress:
Thanks, @froschdesign!

PR #196, opened 2019-10-23
Update head-script.md

rcapile:
- [X] Is this related to documentation?
 typo: Missing `>` at the example

### DIFF
--- a/docs/book/helpers/head-script.md
+++ b/docs/book/helpers/head-script.md
@@ -46,7 +46,7 @@ to use in the element.
 >
 > ```php
 > // jquery template
-> $template = '<div class="book"{{:title}}</div';
+> $template = '<div class="book">{{:title}}</div>';
 >
 > $this->headScript()->appendScript(
 >     $template,


webimpress:
Thanks, @rcapile!

rcapile:
sorry I made things worse.

The documentation uses `>`  to create a quote. It works fine in GitHub, but breaks completely at zf doc site.

https://docs.zendframework.com/zend-view/helpers/head-script/#headscript

I looking up if this a bug from mkdocs or there is something missing

webimpress:
ping @froschdesign 

what do you think about it?

froschdesign:
@rcapile 
> it seems that fenced code blocks inside a quote does not work well with mkdocs

This is not a problem of MkDocs: https://python-markdown.github.io/extensions/fenced_code_blocks/#syntax

rcapile:
I just find out this is a known bug with no fix in sight 

https://github.com/Python-Markdown/markdown/issues/53

froschdesign:
@rcapile 
> I just find out this is a known bug with no fix in sight

Unfortunately, it's not supported and not a bug.

webimpress:
@froschdesign do you think we can solve it somehow by separate script in our template?
As we are doing couple other things - [table_fix_pipes.php](https://github.com/zendframework/zf-mkdoc-theme/blob/master/table_fix_pipes.php), [table_responsive.php](https://github.com/zendframework/zf-mkdoc-theme/blob/master/table_responsive.php), etc...

froschdesign:
@webimpress 
As workaround we can replace the backticks with the standard Markdown for code blocks (4 spaces). Then we can rewrite the HTML from:

```html
<pre class="codehilite line-numbers"><code>…</code></pre>
```

to

```html
<pre class="codehilite line-numbers language-php"><code>…</code></pre>
```

Then the code can be read but without colour.

webimpress:
@rcapile all is working fine now :) We have made some changes in the template and fenced code blocks are working correctly. Thanks!

